PR #9, opened 2020-10-17
Add Homepage Title + Text

w-gao:
Problem
=======
As a user, 
I want to see the title text and description,
So that I am aware of the tentative event dates.

[Link to Asana Ticket](https://app.asana.com/0/1197738625416321/1198219556358095)

Solution
========
In order for the hero to be two columns and responsive, I structured the hero view with a flexbox. 

- Also updated `.stylelintrc.js` so that we can nest more selectors in SCSS. (Not sure if this is the best way; I did it so the linter won't complain)

Change summary:
---------------
* added `responsive-font` mixin from 2020's website
* updated hero view

Steps to Verify:
----------------
* Check this branch out and start a local server
* Verify that the homepage title and text are visible & responsive

Screenshots:
-----------------------
image

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -9,7 +9,7 @@ module.exports = {
       "declaration-empty-line-before": null,
       "declaration-colon-newline-after": null,
       "declaration-no-important": true,
-      "max-nesting-depth": 1,
+      "max-nesting-depth": 3,
       "scss/at-function-pattern": "^[a-z0-9]+(-[a-z0-9]+)*$",
       "scss/at-mixin-pattern": "^[a-z0-9]+(-[a-z0-9]+)*$",
       "scss/dollar-variable-pattern": "^[a-z0-9]+(-[a-z0-9]+)*$",
@@ -24,7 +24,7 @@ module.exports = {
             "Classes must be in BEM form like `MyComponent__element--variant`",
         },
       ],
-      "selector-max-compound-selectors": 1,
+      "selector-max-compound-selectors": 3,
       "selector-max-id": 0,
       "selector-no-qualifying-type": true,
       "sh-waqar/declaration-use-variable": [
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -9,7 +9,7 @@ module.exports = {
       "declaration-empty-line-before": null,
       "declaration-colon-newline-after": null,
       "declaration-no-important": true,
-      "max-nesting-depth": 3,
+      "max-nesting-depth": 2,
       "scss/at-function-pattern": "^[a-z0-9]+(-[a-z0-9]+)*$",
       "scss/at-mixin-pattern": "^[a-z0-9]+(-[a-z0-9]+)*$",
       "scss/dollar-variable-pattern": "^[a-z0-9]+(-[a-z0-9]+)*$",
@@ -24,7 +24,7 @@ module.exports = {
             "Classes must be in BEM form like `MyComponent__element--variant`",
         },
       ],
-      "selector-max-compound-selectors": 3,
+      "selector-max-compound-selectors": 1,
       "selector-max-id": 0,
       "selector-no-qualifying-type": true,
       "sh-waqar/declaration-use-variable": [
--- a/src/stylesheets/_typography.scss
+++ b/src/stylesheets/_typography.scss
@@ -1,59 +1,15 @@
-@use "src/stylesheets/responsiveness";
+@use "~stylesheets/responsiveness";
 
 $font-family-base: "Montserrat", "sans-serif";
 $font-family-italic: "Montserrat Italic", "sans-serif";
 
-// ============================================================================
-// Responsive Fonts
-// ============================================================================
-
-///
-/// Viewport sized typography with minimum and maximum values
-///
-/// @author Eduardo Boucas (@eduardoboucas)
-///
-/// @param {Number}   $responsive  - Viewport-based size
-/// @param {Number}   $min         - Minimum font size (px)
-/// @param {Number}   $max         - Maximum font size (px)
-///                                  (optional)
-/// @param {Number}   $fallback    - Fallback for viewport-
-///                                  based units (optional)
-///
-/// @example scss - 5vw font size (with 50px fallback),
-///                 minumum of 35px and maximum of 150px
-///  @include responsive-font(5vw, 35px, 150px, 50px);
-///
-@mixin responsive-font($responsive, $min, $max: false, $fallback: false) {
-  $responsive-unitless: $responsive / ($responsive - $responsive + 1);
-  $dimension: if(unit($responsive) == "vh", "height", "width");
-  $min-breakpoint: $min / $responsive-unitless * 100;
-
-  @media (max-#{$dimension}: #{$min-breakpoint}) {
-    font-size: $min;
-  }
-
-  @if $max {
-    $max-breakpoint: $max / $responsive-unitless * 100;
-
-    @media (min-#{$dimension}: #{$max-breakpoint}) {
-      font-size: $max;
-    }
-  }
-
-  @if $fallback {
-    font-size: $fallback;
-  }
-
-  font-size: $responsive;
-}
-
-@mixin h1 {
+@mixin title-text {
   margin: 0;
   color: var(--white);
   font-family: $font-family-italic;
   font-weight: normal;
 
-  @include responsive-font($responsive: 4vw, $min: 32px, $max: 50px);
+  @include responsiveness.font($responsive: 4vw, $min: 32px, $max: 50px);
   line-height: 50px;
 
   @include responsiveness.respond(tab-port) {
@@ -61,9 +17,9 @@ $font-family-italic: "Montserrat Italic", "sans-serif";
   }
 }
 
-@mixin title-text {
+@mixin description {
   color: var(--white);
   font-family: $font-family-base;
 
-  @include responsive-font($responsive: 2.4vw, $min: 12px, $max: 18px);
+  @include responsiveness.font($responsive: 2.4vw, $min: 12px, $max: 18px);
 }
--- a/src/stylesheets/_typography.scss
+++ b/src/stylesheets/_typography.scss
@@ -1,14 +1,69 @@
+@use "src/stylesheets/responsiveness";
+
 $font-family-base: "Montserrat", "sans-serif";
 $font-family-italic: "Montserrat Italic", "sans-serif";
 
-@mixin major-title {
-  font-family: $font-family-italic;
-  font-size: 50px;
-  color: var(--wedgewood);
+// ============================================================================
+// Responsive Fonts
+// ============================================================================
+
+///
+/// Viewport sized typography with minimum and maximum values
+///
+/// @author Eduardo Boucas (@eduardoboucas)
+///
+/// @param {Number}   $responsive  - Viewport-based size
+/// @param {Number}   $min         - Minimum font size (px)
+/// @param {Number}   $max         - Maximum font size (px)
+///                                  (optional)
+/// @param {Number}   $fallback    - Fallback for viewport-
+///                                  based units (optional)
+///
+/// @example scss - 5vw font size (with 50px fallback),
+///                 minumum of 35px and maximum of 150px
+///  @include responsive-font(5vw, 35px, 150px, 50px);
+///
+@mixin responsive-font($responsive, $min, $max: false, $fallback: false) {
+  $responsive-unitless: $responsive / ($responsive - $responsive + 1);
+  $dimension: if(unit($responsive) == "vh", "height", "width");
+  $min-breakpoint: $min / $responsive-unitless * 100;
+
+  @media (max-#{$dimension}: #{$min-breakpoint}) {
+    font-size: $min;
+  }
+
+  @if $max {
+    $max-breakpoint: $max / $responsive-unitless * 100;
+
+    @media (min-#{$dimension}: #{$max-breakpoint}) {
+      font-size: $max;
+    }
+  }
+
+  @if $fallback {
+    font-size: $fallback;
+  }
+
+  font-size: $responsive;
 }
 
-@mixin title-info {
-  font-family: $font-family-italic;
-  font-size: 50px;
+@mixin h1 {
+  margin: 0;
   color: var(--white);
+  font-family: $font-family-italic;
+  font-weight: normal;
+
+  @include responsive-font($responsive: 4vw, $min: 32px, $max: 50px);
+  line-height: 50px;
+
+  @include responsiveness.respond(tab-port) {
+    line-height: 40px;
+  }
+}
+
+@mixin title-text {
+  color: var(--white);
+  font-family: $font-family-base;
+
+  @include responsive-font($responsive: 2.4vw, $min: 12px, $max: 18px);
 }
--- a/src/stylesheets/_typography.scss
+++ b/src/stylesheets/_typography.scss
@@ -7,6 +7,7 @@ $font-family-italic: "Montserrat Italic", "sans-serif";
   margin: 0;
   color: var(--white);
   font-family: $font-family-italic;
+  font-style: italic;
   font-weight: normal;
 
   @include responsiveness.font($responsive: 4vw, $min: 32px, $max: 50px);
--- a/src/stylesheets/_typography.scss
+++ b/src/stylesheets/_typography.scss
@@ -1,7 +1,7 @@
 @use "~stylesheets/responsiveness";
 
-$font-family-base: "Montserrat", "sans-serif";
-$font-family-italic: "Montserrat Italic", "sans-serif";
+$font-family-base: "Montserrat", sans-serif;
+$font-family-italic: "Montserrat Italic", sans-serif;
 
 @mixin title-text {
   margin: 0;
--- a/src/views/Hero/Hero.scss
+++ b/src/views/Hero/Hero.scss
@@ -1,5 +1,5 @@
-@use "src/stylesheets/typography";
-@use "src/stylesheets/responsiveness";
+@use "~stylesheets/typography";
+@use "~stylesheets/responsiveness";
 
 .HeroView {
   display: flex;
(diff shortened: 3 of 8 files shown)


src/views/Hero/Hero.scss
```diff
@@ -1,11 +1,56 @@
...
+        color: var(--wedgewood__dark);
+      }
+    }
+
+    > p {
+      max-width: 375px;
+      margin: 2em 0;
+
+      @include typography.title-text;
+    }
```
ilopezro, line 54: Lets stray away from doing this. Use classnames instead!

src/views/Hero/index.view.tsx
```diff
@@ -16,11 +16,19 @@ const HeroView: React.FC<HeroProps> = ({ pageName, title, description }) => {
...
+        </div>
+        <div className="HeroView__text">
+          <h1>CruzHacks 2021 /</h1>
+          <h1>
+            <span>{title}</span>
+          </h1>
+          <p>{description}</p>
+          <EmailSubscriptionForm />
+        </div>
+      </div>
```
ilopezro, line 31: We should give the `h1` and `p` tags classnames. I am also not sure if we want to be nesting `p` tags within `h1` tags; rather, we can nest the `p` tag within a `div` instead. 

```suggestion
        <div className="HeroView__text">
          <h1 className="HeroView__titleText>CruzHacks 2021 /</h1>
          <h1 className="HeroView__titleTextTwo>{title}</h1>
          <p className="HeroView__description">{description}</p>
          <EmailSubscriptionForm />
        </div>
      </div>
```

src/stylesheets/_typography.scss
```diff
@@ -1,14 +1,69 @@
...
+      font-size: $max;
+    }
+  }
+
+  @if $fallback {
+    font-size: $fallback;
+  }
+
+  font-size: $responsive;
 }
```
ilopezro, line 48: Wondering if this would actually go under `_responsiveness.scss` instead of this file. It would make this file much cleaner.

src/views/Hero/Hero.scss
w-gao: Ahh, oops! My bad. 

Although, how would you go about making this nest only once? 

```css
.HeroView {
  &__container {
    // ...
    @include responsiveness.respond(phone) {
      // ...
    }
  }
```

src/stylesheets/_typography.scss
```diff
@@ -1,14 +1,69 @@
+@use "src/stylesheets/responsiveness";
```
ilopezro, line 1: Remember that the root is now the `/src` folder so all the imports will happen from that folder. 

```suggestion
@use "~stylesheets/responsiveness";
```

src/views/Hero/Hero.scss
ilopezro: That's a really good point! How about we increate the nesting to two depth instead of 3?

src/stylesheets/_typography.scss
```diff
@@ -1,14 +1,69 @@
...
+
+  @if $fallback {
+    font-size: $fallback;
+  }
+
+  font-size: $responsive;
 }
 
-@mixin title-info {
+@mixin h1 {
```
ilopezro, line 50: Let's stick with giving mixins more meaningful names. I am not entirely sure what this mixin in used for.

ilopezro (dismissed):
Looks really good overall! Take a look at the comments I left 👍

.stylelintrc.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
             "Classes must be in BEM form like `MyComponent__element--variant`",
         },
       ],
-      "selector-max-compound-selectors": 3,
+      "selector-max-compound-selectors": 1,
```
ilopezro: This should still be 0.

src/stylesheets/_typography.scss
w-gao: I agree. Is something like `description` good enough? This should be something that can be used in many places.

ilopezro: That works for me for now!

src/views/Hero/index.view.tsx
```diff
@@ -16,11 +16,17 @@ const HeroView: React.FC<HeroProps> = ({ pageName, title, description }) => {
...
-      Hello from HeroView
-      {title}
-      {description}
+      <div className="HeroView__container">
+        <div className="HeroView__visual">
+          <PostcardStack pageName={pageName} />
+        </div>
+        <div className="HeroView__text">
+          <h1 className="HeroView__titleText">CruzHacks 2021 /</h1>
+          <h1 className="HeroView__titleTextTwo">{title}</h1>
```
ilopezro, line 25: Awesome! If we're going with this approach, I'd suggest adding a modifier to that `titleText` class that we have up already. 

```suggestion
          <h1 className="HeroView__titleText--secondary">{title}</h1>
```

In the styling we can nest the following together: 
```scss
&__titleText {
    @include typography.title-text;
    &--secondary {
           color: var(--wedgewood__dark);
    }
  }
```

src/views/Hero/index.view.tsx
```diff
@@ -16,11 +16,17 @@ const HeroView: React.FC<HeroProps> = ({ pageName, title, description }) => {
...
-      <PostcardStack pageName={pageName} />
-      <EmailSubscriptionForm />
-      Hello from HeroView
-      {title}
-      {description}
+      <div className="HeroView__container">
+        <div className="HeroView__visual">
+          <PostcardStack pageName={pageName} />
+        </div>
+        <div className="HeroView__text">
```
ilopezro, line 23: Super nit-picky, but suggest changing the class name to `HeroView__textContainer` just so we can tell it apart from something else that might be called `text`. This is more of a container.

src/views/Hero/index.view.tsx
w-gao: Yeah, the properties don't inherit. I think `HeroView__titleText--secondary` looks nicer though so I'll change the class name to that.

src/stylesheets/_typography.scss
```diff
@@ -1,14 +1,26 @@
...
 
-@mixin major-title {
+@mixin title-text {
+  margin: 0;
+  color: var(--white);
   font-family: $font-family-italic;
-  font-size: 50px;
-  color: var(--wedgewood);
+  font-style: italic;
+  font-weight: normal;
```
ilopezro, line 11: If we want normal, italicized text, we can add the fonts for these. I can create a PR and push it against yours if you're not sure how to do this.

w-gao: Go ahead! I'm not entirely sure how to do this.

ilopezro: I took a look at the code and we can actually do without those two lines of code. We are calling the italic font with the 500 `font-weight` and that's exactly what the design calls for.

w-gao: Actually no. `font-style` is necessary for the font to display correctly on iOS (Safari and Chrome) for reasons that I’m not entirely sure. 

My understanding is that `font-style` in `@font-face` specifies the style supported by that font, which is only italics in this case. So we probably have to change the default font style for h1 to match. 

See [this](https://stackoverflow.com/questions/23731492/font-face-does-not-work-properly-for-italic-bold-fonts) questionable Stack overflow answer for reference.

src/views/Hero/index.view.tsx
```diff
@@ -16,11 +16,17 @@ const HeroView: React.FC<HeroProps> = ({ pageName, title, description }) => {
...
-      {title}
-      {description}
+      <div className="HeroView__container">
+        <div className="HeroView__visual">
+          <PostcardStack pageName={pageName} />
+        </div>
+        <div className="HeroView__textContainer">
+          <h1 className="HeroView__titleText">CruzHacks 2021 /</h1>
+          <h1 className="HeroView__titleText--secondary">{title}</h1>
+          <p className="HeroView__description">{description}</p>
```
tim-nguyen-cs, line 26: Would it also be advisable to change `h1` and `p` to `div` to remove any default styling that we might have to repetitively overwrite?

src/views/Hero/Hero.scss
```diff
@@ -1,11 +1,57 @@
...
+    @include responsiveness.respond(phone) {
+      margin: 1em;
+    }
+  }
+
+  &__visual {
+    // postcard styling
+    max-width: 500px;
+    margin: 0 auto;
+  }
```
tim-nguyen-cs, line 36: Is this class also necessary when the `PostcardStack` comprises its own component?

ilopezro: This might be necessary because remember that `PostcardStack` will only style the component itself independent of where it goes on the hero page.

src/stylesheets/_typography.scss
ilopezro: To piggy back [Tim's comment](https://github.com/CruzHacks/cruzhacks-2021-website/pull/9#discussion_r507198544), I'm wondering if this might be due to some underlying default styling by the `h1` and `p` tags.

src/views/Hero/Hero.scss
```diff
@@ -1,11 +1,57 @@
+@use "~stylesheets/typography";
+@use "~stylesheets/responsiveness";
+
 .HeroView {
-  height: 100%;
+  display: flex;
+  overflow: hidden;
+  justify-content: center;
+  min-height: 100%;
```
tim-nguyen-cs, line 8: When I inspect the code this style is actually overriden by `min-height: 90%` - do we still need to include this?

src/views/Hero/Hero.scss
```diff
@@ -1,11 +1,57 @@
...
-    height: 90%;
+    min-height: 90%;
     border-bottom-left-radius: 50px;
     border-bottom-right-radius: 50px;
     box-shadow: 0 10px 16px rgba(0, 0, 0, 0.25);
   }
+
+  &__container {
+    display: flex;
+    flex-direction: row;
```
tim-nguyen-cs, line 20: `flex-direction: row` is declared by default, so I'm not sure if we need this.

src/views/Hero/Hero.scss
```diff
@@ -1,11 +1,57 @@
...
+    flex-wrap: wrap;
+    justify-content: center;
```
tim-nguyen-cs, line 22: It looks like each child element within this container is being sized to-scale (i.e. the widths and margins of the two child elements fill up the entire container). In this case, would we still need this property?

src/views/Hero/index.view.tsx
w-gao: Perhaps. Would it make the TSX file less readable if we don't use the heading and `p` tags for text? To me `div` tags are more for containers but that's a personal preference.

src/stylesheets/_typography.scss
w-gao: That doesn't seem to be the case. The font is still not showing up correctly on mobile with the `div` tag.

src/views/Hero/index.view.tsx
mocklee: We could also take advantage of the `h1, p, ...` elements by redeclaring their default styles in one of the base stylesheets!

mocklee: Another route would be to completely make text elements generic with `span`, although that's a pretty big compromise to HTML readability

src/stylesheets/_typography.scss
mocklee: if the font isn't being rendered at all, it probably points to an issue with the `font-family` reference. 

what's the relevance of appending `"sans-serif"` to the variable declaration here?

```
$font-family-base: "Montserrat", "sans-serif";
$font-family-italic: "Montserrat Italic", "sans-serif";
```

src/views/Hero/index.view.tsx
tim-nguyen-cs: I think that's a great recommendation - we can forgo class names in exchange for general header formatting.

src/views/Hero/Hero.scss
tim-nguyen-cs: Ahh, that makes sense - we'll have to re-ensure that the `PostcardStack` composes only of itself rather than the container surounding it.

src/stylesheets/_typography.scss
ilopezro: I added those in, but we can take those out. It was the name of the font on the site that I used to download the fonts.

w-gao: I might've worded it weird. The font is being rendered, but it's not Montserrat italic on mobile. It works on desktop though. 

Also, I think `"sans-serif"` should be without quotes?

src/views/Hero/index.view.tsx
ilopezro: @tim-nguyen-cs we want to stay organized with class names. We don't want to eliminate classmates. What we can do is add a general mixin for an `h1`, `p`, etc tags and include that mixin for each div.

w-gao: @tim-nguyen-cs I think @ilopezro wants to use class names though (https://github.com/CruzHacks/cruzhacks-2021-website/pull/9#discussion_r506986915), unless I'm misunderstanding your suggestion?

src/stylesheets/_typography.scss
mocklee: yep, `sans-serif` is a generic/fallback font-family, which is a sort of baseline that CSS sees as an actual property (like colors) 
https://developer.mozilla.org/en-US/docs/Web/CSS/font-family

src/views/Hero/index.view.tsx
mocklee: @ilopezro I don't think we need to introduce unnecessary abstraction if we can avoid it— mixins are useful, but if we can select the HTML elements themselves why bother? I do agree that CSS classes are important & useful for granular styling, but if we can establish baseline style formatting let's do it.

ilopezro: Especially in the `Hero` component, we don't really need the usage of mixins that much. We will be reusing hero component throughout our views. Let's stick with class names + defining everything at the Hero level since I don't see us reusing those outside of the hero component in order to avoid over abstraction! how does that sound y'all? I do still think we should change the `h1` tags into `div` tags and style the text under a `span` tag (or simply under the `div` tag), but that's just me.

mocklee: @ilopezro divs containing spans would get the job done, but it blows up the HTML file size if we introduce that redundancy. Plus it seems that using traditional HTML element references promotes readability (let's keep this source code interesting so it can be workshop material)

src/views/Hero/Hero.scss
w-gao: We probably don't want to remove it here since it's a fallback height. If all Hero components have the same height then we can just set it here once and not override it. 

@ilopezro Thoughts?

src/views/Hero/index.view.tsx
w-gao: Have we reached a consensus here?

src/views/Hero/Hero.scss
ilopezro: Correct that's the intended purpose of having the different class names within the hero page as well to prepare for the future views.

src/views/Hero/index.view.tsx
tim-nguyen-cs: Just putting in my two cents, I am recommending these two approaches:

1. `div` tags with class names (alternatively also `span` tags for inline styling)
2. `h*` and `p` tags without class names (because the overlap would be redundant/superfluous)

This is just based on the development experience I've been encompassing, but let me know what the "law of the land" says.

ilopezro: I like the idea of using `h*` and `p` tags without class names but the issue with that is that the CSS will get hard to follow and readability will be low. Also, by adding defaulting values for these tags as well, we are making a big assumption that all those tags will be rendered the same and I don't like that approach. I'd rather just define the `div` to what I want it to look like and using `span` or just plain text within the div itself.

mocklee: why not both & have classnames + simple elements? we can reset the default styles for established HTML elements to our own design pattern using base stylesheets & use granular classnames as well since CSS cascades

ilopezro: I think that would defeat the purpose of having CSS files within each component and view if we're defining basic elements at the base level. I think let's stick with class names since this is the only place where I foresee we're using these specific styles. These are hero specific so I don't see why we should redefine these styles higher up in the tree. Let's keep these styles component specific w/ class names. 

Again, my suggestion here would be `div` and a `p` nested within the divs but `h1` also works for me. I usually don't use the header tags and nest a `span` or `p` within a `div`.

mocklee: here's a demo of how cascading base HTML elements into `div` containers can help us cut some abstraction: https://codepen.io/mocklee/pen/wvWzxzV

ilopezro: Ok i see that. I personally don't like this approach because to reiterate, we'd be making a huge assumption about our tags if we move it further up the tree. The further down you to you have to set and unset props and that seem unnecessary. If there is code that we can refactor, we can should create mixins instead. We should make full use of the SCSS magic.

mocklee: updated the demo to demonstrate nested classnames. I'd say that refactors w/ adding granular classnames later on are tolerable, as long as there aren't too many distinct changes. 

also, it seems that mixins are being siloed to serve more as functions— [the Sass standard also has @extend available, which is syntactical sugar for short mixin declaration](https://sass-lang.com/documentation/at-rules/extend).

ilopezro: Yeah I noticed the updated Codepen environment you've got up. If we go with pure classnames, there is no need for that mixin to unset the properties. That becomes redundant when you redefine them inside the another component. I wouldn't want to get too much technical debt if we over refactor and then later on find out we're doubling up on code because we've over refactored at the beginning. Let's stick with classnames for now. When there is a place to refactor, lets go for it!

mocklee: [SCSS psuedoclasses](https://css-tricks.com/the-sass-ampersand/) only run one level deep, don't overthink nesting! In general, CSS inheritance isn't hard; but nuanced— it depends on both local pseudoclasses & general DOM (HTML) tree structure. if classnames make it easier, we'll go with that.

if `%moreStyles` seems redundant, please demo an improvement!

ilopezro (approved):
This looks really good overall man! Great work 🎉

PR #10, opened 2020-10-18
Using Typescrip to the Max

ilopezro:
Problem
=======
Typescript itself provides type checking, but I accidentally used the `type` and `PropTypes` combination to type check the props, which is redundant because Typescript does that automatically with the usage of `interface`, so this PR addressed that issue.  

Also, we decided to put all `@types` as dev dependencies. These are only used for type checking and never get imported into our codebase whatsoever. 

Steps to Verify:
----------------
1. Verify that `prop-types` and `@types/prop-types` is no longer in the codebase 
1. All `@types/**` dependencies are now dev dependencies 

Screenshots (optional):
-----------------------


ilopezro:
These changes were incorporated in PR #14, so I will close this PR up and open a new PR on a separate branch to introduce the testing suite we will be using.

PR #8, opened 2020-10-16
Addition of Hero Component

ilopezro:
Problem
=======
As a developer, 
I want to see a boilerplated hero component, 
So I can focus on implementing my specific component 

[Link to Asana Ticket](https://app.asana.com/0/1197738625416321/1198489500602053/f)

Solution
========
In order to refactor the hero component for our future postcard components, we created a `Hero Component` which take in the following props: `pageName`, `title`, and `description`. These props will be used within the `Hero` component to render the hero for each specific view `Schedule`, `FAQ`, etc. In order to do some further refactoring, we have added a `pageName` prop to `PostcardStack` component to have it render the given postcard in any view. 

Change summary:
---------------
* added hero view and images folder 3ec86c0
* implemented background c24a4f1
* added additional props to hero view 76bf456

Steps to Verify:
----------------
1. Check this branch out and start a local server 
1. Verify that the gradient background is visible & responsive 

Screenshots:
-----------------------
Screen Shot 2020-10-15 at 10 06 00 PM

PR #11, opened 2020-10-22
Adding Email Subscription Component

tim-nguyen-cs:
Problem
=======
As a user,
I would like to subscribe to CruzHacks' newsletter
to receive timely updates about this years hackathon!

[Link to Asana Ticket](https://app.asana.com/0/1197738625416321/1197738625416326/f)

Solution
========
To create an email subscription component, we first create a responsive element based on the provided Figma design, which only includes an input field and submission button. The two elements are separated by a single right margin on `input` and wraps left-aligned using `flex-wrap: wrap`. The `input` component itself then resizes according to the width of the current viewports.

To handle user input, the `input`'s `onCapture` event is handled, which updates a local state variable corresponding to the user's current input. We also capture its `keyDown` event to submit their email whenever they input the "Enter" key. On the `button` component, we then submit their email when the button is clicked.

To create an asynchronous request, Axios is used to send an HTTP request to the designated Mailchimp endpoint and a feedback message is appropriately bound and displayed to the user using the `showFeedback` conditional.

Change summary:
---------------
* Added more accurate `Arrow` SVG and updated minimum breakpoint to include iPhone X 475c3fd0575617db4287fd9542c4f65d1387fa01
* Finished responsive layout e9ed21da4223269e0b97f8396e712bd1ace53067
* Handled user input  e31bb975dc125051302aa380fe688c7ef09e4c7e
* Configured Axios request 94c3d81be87ff24c139329ee9c02820c72cb6d9e

Steps to Verify:
----------------
1. Checkout branch using `git checkout feature/email-subscription`
3. Open Chrome Developer Tools using `CMD/CTRL + SHIFT + I` to verify that email subscription component is responsive
4. Submit email in form to verify that appropriate response message is returned


src/components/EmailSubscription/EmailSubscription.scss
```diff
@@ -1,3 +1,40 @@
...
 
-/* Delete these lines when you begin development */
+.EmailSubscription {
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  width: 100%;
+
+  .SearchBox {
+    font-family: "Montserrat Italic";
```
tim-nguyen-cs, line 11: I know we have designated fonts in `_fonts.scss` and that we were talking to Will about this, but what's the right way to import fonts from that file?

ilopezro: That works! Take a look at how it's done in the `_fonts.scss` file. We defined a constant at the top and use that constant throughout the file.

src/components/EmailSubscription/EmailSubscription.scss
```diff
@@ -1,3 +1,40 @@
...
+
+  .InputButton {
+    background: var(--wedgewood);
+    border-radius: 50%;
+    padding: var(--medium-spacer);
+    border: none;
+  }
+  .FeedbackMessage {
+    @include typography.title-info; // Will change to 'description' after Will's PR
+  }
```
ilopezro, line 39: Let's use the BEM Naming Conventions for these. I looked at the TS file and it looks like the nested props all live under `EmailSubscription`. Let's rename them to `EmailSubscription__inputButton`, etc. We can keep the nestings here and just make use of the ampersand convention that SCSS provides. `&__inputButton`

tim-nguyen-cs: I'm not sure what you mean by using the ampersand convention, because each subcomponent doesn't inherit the default `EmailSubscription` class, but take a look at the revisions that I made, and let me know if there's anything else that I need to do.

src/components/EmailSubscription/EmailSubscription.scss
```diff
@@ -1,3 +1,41 @@
...
+@use "~stylesheets/typography";
 
-/* Delete these lines when you begin development */
+.EmailSubscription {
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  width: 100%;
+
+  .EmailSubscription__searchBox {
```
ilopezro, line 10:
```suggestion
  .&__searchBox {
```

tim-nguyen-cs: Ohh, that's what you meant!

tim-nguyen-cs: @ilopezro - any recommendations for circumventing the `selector-class-pattern`  StyleLint error: `Classes must be in BEM form like `MyComponent__element--variant``

If we include the ampersand in the class name, it doesn't match the convention.

ilopezro: I'll take a look at it!

.stylelintrc.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
       "declaration-empty-line-before": null,
       "declaration-colon-newline-after": null,
       "declaration-no-important": true,
-      "max-nesting-depth": 1,
+      "max-nesting-depth": 3,
```
ilopezro, line 12: nesting depth is 2, max compound selector should go back to 1. We want to enforce these rules and not change them to our convenience.

.prettierrc.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
     "tabWidth": 2,
-    "useTabs": false
+    "useTabs": false,
+    "endOfLine": "auto"
```
ilopezro, line 4: Same here. Please change these back.

src/components/EmailSubscription/EmailSubscription.scss
ilopezro: Yes, so you have to 
```suggestion
  &__searchBox {
  }
  &__inputButton {
  }
  &__feedbackButton {
  }
```

.prettierrc.js
tim-nguyen-cs: Ahh, this was partially my fault - I had default line endings as `CRLF` instead, so a bunch of errors were showing up.

src/components/EmailSubscription/EmailSubscription.scss
tim-nguyen-cs: But that's not how modifying classes work with ampersands... ampersands are only applicable if that element also inherits its respective parent class. Taking a look at the screenshot below, we can see the `input` and `button` elements do not inherit their parent `EmailSubscription` class, so the ampersands will not apply correctly (also tested).

Can you clarify you perspective?

image

ilopezro: Sounds good. Then let's move away from nesting if they don't inherit parent props. Instead of it all being under the `.EmailComponent` tag let's put them outside

tim-nguyen-cs: A little confused by what you said there - are you talking about removing the `EmailSubscription` class/wrapper-element?

ilopezro: By having everything nested within the `.EmailComponent` class, we are assuming that it's children inherit all of their props. Let's move whatever doesn't belong inside `.EmailComponent` outside of that class. So we can see a structure like 
```scss
.EmailComponet{...props}
.EmailComponent__inputButton {...props} 
```

tim-nguyen-cs: Okay, so just to understand you correctly - do you want me to extrapolate `searchBox` and `inputButton` into their own components? Wouldn't that be too verbose of an extrapolation - they're just HTML components.

ilopezro: https://github.com/CruzHacks/cruzhacks-2021-website/blob/8c0f58a4e70459d732df92953cf211e1e620c22c/src/components/EmailSubscription/EmailSubscription.scss#L4-L41

ilopezro: This is what I was doing before you suggested [nesting them](https://github.com/CruzHacks/cruzhacks-2021-website/pull/8#discussion_r506146642). So if they do not inherit any of the props, just have it go outside of the `.EmailSubscription` component and have it be it's own component. This will help with readability.

tim-nguyen-cs: Sorry, I misunderstood you with my last comment. It's not too verbose and works alright, but I wouldn't say it's `its own component` - it's just not nested, which is different in my perspective.

src/components/EmailSubscription/EmailSubscription.scss
```diff
@@ -1,3 +1,45 @@
...
+@use "~stylesheets/typography";
 
-/* Delete these lines when you begin development */
+.EmailSubscription {
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  width: 100%;
+
+  &__searchBox {
```
w-gao, line 10: I'm seeing this border on focus. Should be able to get rid of it with something like

```scss
&:focus {
  outline: none;
}
```
image

w-gao (commented):
Overall, this looks great! You can probably constrain the width from the Hero view since it looks pretty large. I don't think there's any styling applied to the EmailSubscription component right now.

ilopezro (changes requested):
Also, please fix this issue happening on your branch. I am not 100% sure what is happening or why it's happening, but the components are stacking on a normal laptop view, which shouldn't be happening at all.

Screen Shot 2020-10-26 at 6 05 44 PM

mocklee:
@ilopezro the early flex-wrap is likely because of the postcard... its responsiveness might need some slight reworking to gradually decrease postcard size, so it's not constant width. I'll rework those responsiveness updates once Tim's branch is finalized

src/components/EmailSubscription/EmailSubscription.scss
mocklee: I wouldn't remove input outlines; they're an accessibility feature https://www.a11yproject.com/posts/2013-01-25-never-remove-css-outlines/

package.json
```diff
@@ -6,6 +6,14 @@
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.3.2",
     "@testing-library/user-event": "^7.1.2",
+    "@types/classnames": "^2.2.10",
+    "@types/jest": "^26.0.14",
+    "@types/node": "^14.11.1",
+    "@types/prop-types": "^15.7.3",
+    "@types/react": "^16.9.49",
+    "@types/react-dom": "^16.9.8",
+    "@types/react-router-dom": "^5.1.6",
```
ilopezro, line 15: There seems to be an issue with the merge with main. Please remove these `@types` from the dependencies and keep them in your `devDependencies`. I rebased main with your branch locally it fixed the issue! Whenever you're fixing merge conflicts, please be aware of what the changes are.

src/components/EmailSubscription/index.view.tsx
```diff
@@ -1,9 +1,67 @@
...
+  const handleKey = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.key === "Enter") {
+      submitEmail();
+    }
+  };
+
+  const submitEmail = () => {
+    axios
+      .post(
+        "https://us-central1-cruzhacks-4a899.cloudfunctions.net/subscribe",
```
mocklee, line 25: Let's try to contain implementation constants (like API endpoints) as environment secrets when possible

mocklee (changes requested):
backtracking a bit on the responsiveness issue @ilopezro raised— it seems that the early `flex-wrap` is indeed mostly being caused by the width of the email input; that component takes up the width of its entire parent. Try adding a `max-width` to the email input container, rather than `width: 100%`! 

Maybe the `max-width` should go on the parent container instead, it will take some tweaking to converge on properties that work.

one more gripe— there isn't any immediate user feedback when the send button is pressed. Maybe change the state of the input placeholder on submission w/ a loading message before vanishing the email component completely. Also, consider only vanishing the component on successful submission, and leaving an error message in the placeholder text when submit fails. The 2020 email component has a good example of this kind of UX (although the input is never locked)!

src/components/EmailSubscription/index.view.tsx
```diff
@@ -1,9 +1,67 @@
...
+        { email: userEmail }
+      )
+      .then((response: AxiosResponse) => {
+        setShowFeedback(true);
+        setFeedbackMessage(response.data.message);
+      })
+      .catch(() => {
+        setShowFeedback(true);
+        setFeedbackMessage(
+          "Oh no! Your submission looks like it failed - please try your request again!"
```
mocklee, line 35: this error message may be too long— we will need to think about wrapping this text on mobile

ilopezro:
To piggy back @mocklee 's comments, the component itself looks really good. I'd suggest now adding the styling to the `Hero` view to make it all come together!

src/components/EmailSubscription/EmailSubscription.scss
tim-nguyen-cs: The Figma design calls for no outline on the input box. What's the call here?

src/components/EmailSubscription/index.view.tsx
tim-nguyen-cs: Can you clarify again how this is done or point me toward resources that help?

ilopezro: Working on this. Will get back to you ASAP!

src/stylesheets/_responsiveness.scss
```diff
@@ -1,6 +1,6 @@
 @mixin respond($breakpoint) {
   @if ($breakpoint == smallest) {
-    @media (max-width: 422px) {
```
ilopezro, line 3: I'd stray away from changing these constants. Please, if any, try and adjust the component using a `min-width` or a `max-width`.

src/components/EmailSubscription/index.view.tsx
```diff
@@ -1,9 +1,75 @@
...
+    }
+  };
+
+  const submitEmail = () => {
+    setShowFeedback(true);
+    setFeedbackMessage("Please wait while your message is being submitted...");
+    setUserEmail("");
+
+    axios
+      .post(process.env.REACT_APP_SUBSCRIBE_ENDPOINT || "", {
```
ilopezro, line 29: I'd create a variable for this and instead of just have the variable value in here. 

```tsx
const subscribeEndpoint = process.env.REACT_APP_SUBSCRIBE_ENDPOINT || ""
...
axios.post(subscribeEndpoint)...
```

This will allow for cleaner and more readable code.

src/stylesheets/_responsiveness.scss
tim-nguyen-cs: I mentioned this above, but I changed this to accommodate for smaller phone resolutions, such as the iPhone X - I can change this back if need be.

I'm also confused about what you mean by your second comment. I've checked the responsive of the component itself, and I think it looks fine. Let me know if there's something specific (per screen resolution) you want me to take a look at!

(In short, `max-width` on `EmailSubscription` can work, but it might not be necessary.)

src/components/EmailSubscription/EmailSubscription.scss
tim-nguyen-cs: @mocklee Bump!

src/stylesheets/_responsiveness.scss
ilopezro: When you import the component here: 
https://github.com/CruzHacks/cruzhacks-2021-website/blob/9b61e7d46b04f92d808f3bea3949577917232fb8/src/views/Hero/index.view.tsx#L27-L32
You should wrap this within its own container and style that to fit within the larger view.

tim-nguyen-cs: Little confused about the difference, but I implemented it regardless! Let me know if we should change back the smallest resolution also, but take a look at it!

ilopezro: Because remember that each component is styled by itself in the `components` directory, but we need to integrate the component to each view and that's where we fix widths and all that good stuff.

ilopezro: As to your earlier point about changing the dimension of this, then I think you should be ok! I'm not sure how we came up with the breakpoints last year so this change is ok!

src/components/EmailSubscription/index.view.tsx
ilopezro: Bump in this @tim-nguyen-cs

src/components/EmailSubscription/index.view.tsx
```diff
@@ -1,9 +1,76 @@
-import React from "react";
+import React, { useState } from "react";
+import axios, { AxiosResponse } from "axios";
+import { ReactComponent as Arrow } from "images/arrow.svg";
```
ilopezro, line 3: These absolute paths looking clean 😤

ilopezro:
I'm being super nit picky, but let's try and work on the responsiveness of the component a bit more? I'm wondering why the SVG looks lopsided in this view (iPhone X). Also, let's try and have the component take up the entire width of that whole div. 

Screen Shot 2020-10-28 at 10 55 20 PM

src/components/EmailSubscription/index.view.tsx
tim-nguyen-cs: Take a look at how I hard-coded the widths on my current revision (pushed) and current iteration (not yet pushed), and let me know what else we should do:
```
  &__feedbackMessage {
    @include typography.description;
    color: var(--wedgewood);
    width: 350px;
    margin-top: var(--extra-small-spacer);

    @include responsiveness.respond(mission) {
      width: min(48vw, 300px);
    }
  }
```

In short, it wraps relative to the search box.

src/components/EmailSubscription/index.view.tsx
tim-nguyen-cs: Resolving as a parenthetical comment!

tim-nguyen-cs:
@ilopezro Take a look at the newest revision! For some reason, I couldn't also get that error you showed to replicate on my end - here's what I currently see:

image

Let me know what else we can do!

mocklee:
@tim-nguyen-cs Looks like the flex elements aren't justified towards the center of their rows— try adding `justify-content: center` on the parent flex containers https://css-tricks.com/almanac/properties/j/justify-content/#demo

tim-nguyen-cs:
@mocklee I added `margin: auto` on the parent container to center align that one element only. If we add justify-content: center` on the entire "text column" container, it'll center align all other elements (unless this is something we want). Let me know if this works (not also that the screenshot above was not the most recent iteration that is currently pushed - that was just to showcase that I couldn't replicate Isai's error).

mocklee (approved):

ilopezro (changes requested):
This LGTM, but let's hold off merging this until #17 gets merge into this branch. I'm trying to work something out with @mocklee

ilopezro (approved):
🚢

src/components/EmailSubscription/index.view.tsx
```diff
@@ -1,9 +1,70 @@
...
+      .then((response: AxiosResponse) => {
+        setFeedbackMessage(response.data.message);
+
+        if (response.status === 200) {
+          setShowInput(false);
+        }
+      })
+      .catch(() => {
+        setFeedbackMessage(
+          "Oh no! We've got an error— please try your request again & contact us if this persists!"
```
ilopezro, line 32: It's a very simple fix! Just add our email here. No need for a hyperlink. 
```suggestion
          "Oh no! We've got an error— please try your request again & contact us at dev@cruzhacks.com if this persists!"
```

PR #15, opened 2020-10-27
Setup Testing Suite

ilopezro:
Problem & Solution
=======
TESTING TESTING TESTING!! This is a critical part of web development and software engineering in general! We always want to test our code. This year we are going to do integration tests using [`jest-puppeteer` ](https://github.com/smooth-code/jest-puppeteer/blob/master/README.md). [Integration tests](https://martinfowler.com/bliki/IntegrationTest.html) are aimed to test the at a more modular level rather than a unit level. Rather than testing components individually, we will test how they work in conjunction with each other. 

I have stripped away the `--passWithNoTest` flag from the `jest` command and we will now start to test our own code at a `views` level. We want to see how all the components work with one another there. 

**NOTE**: In order to run the tests locally, create a build within your repository. You can do this by running `yarn build`. The reason for this is that `jest-puppeteer` spins up a server where it will open up Chromium and run your static files there. It's important to build before you run your tests or else you're probably testing on legacy code. 

Steps to Verify:
----------------
1. Visit the `Checks` tab and navigate to the `js-tests` workflow and you'll see the tests being run. 

Screenshots:
-----------------------
Screen Shot 2020-10-27 at 10 52 39 PM


mocklee (approved):
lit! glad we can actually create common sense test cases w/ `jest-puppeteer`

PR #19, opened 2020-11-01
little fix for email input refactor

mocklee:
Problem
=======
As a new email subscriber, I would like to send a POST request with the actual value of my input, rather than an empty string

Solution
========
Bind email in request with setState() on email input


tim-nguyen-cs (approved):
Freudian slip ~ glad we caught that!

PR #22, opened 2020-11-02
Responsiveness for smaller mobile screens + Add event date

w-gao:
Problem
=======
As a small phone owner, I would like to view the entire content of the site. 

Solution
========
Make site responsive on devices with smaller screens. 

- Remove `max-height` so users can scroll when content doesn't fit on smaller screens. 
- Instead, `align-content` is added so we don't stretch the content when we switch to one column display. 

Screenshots (optional):
-----------------------
Before:
before

After:
after


ilopezro (approved):
This looks awesome!

src/views/Home/index.view.tsx
```diff
@@ -10,7 +10,7 @@ const title = [
     style: "",
   },
   {
-    text: "Coming Soon",
+    text: "January 15-17, 2021",
```
ilopezro, line 13: Added this date in here because we want to be [MLH compliant](https://cruzhacks2021-team.slack.com/archives/C013X5SEH37/p1604359293028800) in order to get that badge!

mocklee: ^ dropped the year off the subheader since we already say it's 2021 in the title

mocklee:
dropped in some responsiveness updates for `tab-land`— the postcard scaling is a little tricky around that breakpoint. everything looks really solid to me, good job y'all!!

PR #55, opened 2020-11-08
Email Icons Changes

ilopezro:
Problem & Solution 
=======
In order to prevent dependency on external links, we can use [Material Design Icons](https://dev.materialdesignicons.com/getting-started/react) React components. I've used their `Icon` component along with their SVG images of the icons used!

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@mdi/js": "^5.8.55",
+    "@mdi/react": "^1.4.0",
     "axios": "^0.21.0",
     "classnames": "^2.2.6",
     "firebase-tools": "^8.15.1",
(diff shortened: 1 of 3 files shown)


tim-nguyen-cs (approved):
Not sure if we are 110% with MatDesign, but looks great to me!

ilopezro:
@tim-nguyen-cs the reason why I used Material Design is because your image source was `https://icongr.am/material/check-circle-outline.svg?size=25&color=4f728e`. So although we were grabbing from [IconGram](https://www.icongr.am/) to grab the Icon, the icons came directly from Material Design.

tim-nguyen-cs:
@ilopezro I understand that. What I'm saying is that I'm not sure what icon library we should be using specifically - it's more so a parenthetical comment versus a comment against your PR. Hope this clarifies it!

ilopezro:
@tim-nguyen-cs ahh ok gotcha! I'm totally okay using this library 👍🏽

PR #79, opened 2020-11-24
new color scheme to lead into upcoming design

mocklee:
Problem
=======
Let's keep the color scheme dark so it'll be consistent with the design! Found these colors using the [Dark Reader](https://chrome.google.com/webstore/detail/dark-reader/eimadpbcbfnmbkopoojfekhnkhdbieeh?hl=en-US) extension on Chrome

Screenshots (optional):
-----------------------
Screenshot from 2020-11-23 21-20-06

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@
 # misc
 .DS_Store
 .dccache
+.eslintcache
 .env.local
 .env.development.local
 .env.test.local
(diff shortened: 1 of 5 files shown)


tim-nguyen-cs (approved):
We might want to adjust the contrast on the placeholder and input box, but looks good to me otherwise!

mocklee:
> We might want to adjust the contrast on the placeholder and input box, but looks good to me otherwise!

spotted the opacity prop & bumped contrast, hopefully this is better
Screenshot from 2020-11-23 22-48-31
:

PR #80, opened 2020-11-24
fix email button & postcard on mobile

mocklee:
Problem
=======
As a mobile user, I find circular buttons favorable over elliptical ones

Solution
========
Unset height unit from `vh` so it doesn't shrink one-dimensionally

Screenshots (optional):
-----------------------
before:
Screenshot from 2020-11-23 23-54-17

after:
Screenshot from 2020-11-23 23-55-21

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -10,7 +10,6 @@
     background: var(--wedgewood);
     border-radius: 15px;
     padding: var(--extra-small-spacer) var(--large-spacer);
-
     border: none;
     cursor: pointer;
   }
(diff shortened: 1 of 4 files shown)


ilopezro (changes requested):
Are these not the same commits that are already in our `main` branch?

mocklee:
> Are these not the same commits that are already in our `main` branch? 

change is on the last commit, histories got carried over when I switched branches around on my local.

mocklee:
rebased!

ilopezro (approved):
This looks amazing dude! 😍 💯

PR #91, opened 2020-12-02
Milestones component.

w-gao:
Problem
=======
As a hacker, 
I want to see CH milestones,
So I can see how CH has done in the past.

[Link to Asana Ticket](https://app.asana.com/0/1197738625416321/1199333062646019)

Solution
========
- In order to create the cyan-red effect for the title texts, I created the `anaglyph-title-text` mixin which can be used by the Missions, Milestones, and Topics components. 
- I also converted the background grid to png so it can be repeated vertically. 
- Each Milestone lives inside its own `Milestones__box` div in a container. Flex wrap is allowed so the milestones can wrap onto multiple lines on smaller screens.

Change summary:
---------------
- Add Milestones component.

Dev-Ops
=======
- [ ] `.env` & `.env.sample` 
- [ ] `webpack.config.js` 
- [ ] GCP Secret Manager
- [ ] Github Secrets 
- [ ] GithubActions Workflow

Steps to Verify:
----------------
1. Checkout branch and start local server.
2. Verify Milestones component is present & responsive in home page.

Screenshots (optional):
-----------------------


src/views/Home/components/Milestones/index.view.tsx
```diff
@@ -1,8 +1,39 @@
...
-  return <div className="Milestones">Hello from MilestonesComponent</div>;
+  return (
+    <div className="Milestones">
+      <h1 className="Milestones__title">Milestones of 2020</h1>
+      <div className="Milestones__container">
+        {Object.entries(milestones).map(([_, milestone]) => (
+          <div className="Milestones__box" key={_}>
+            <h1 className="Milestones__box--top">{milestone.text}</h1>
+            <p className="Milestones__box--bottom">{milestone.title}</p>
+          </div>
```
tim-nguyen-cs, line 32: Should we change `h1` and `p` to `div` to be more consistent throughout the application? 

Out naming convention also associates `--` with variations - is this the correct way to use it here?

src/views/Home/components/Milestones/milestones.scss
```diff
@@ -1,3 +1,59 @@
...
+  margin: 0;
+  min-height: 500px;
+  background: url("~images/components/milestones/background.png") repeat-y, #1a1542;
+  background-size: 100% auto;
+
+  &__title {
+    @include typography.title-text;
+
+    margin: var(--extra-large-spacer) 0;
+    text-shadow: 3px 3px #0085ff, -2px -2px rgba(255, 0, 0, 0.57);
```
tim-nguyen-cs, line 15: We might have to create variables for the text shadows, but I'm not too sure about that.

src/views/Home/components/Milestones/index.view.tsx
```diff
@@ -1,8 +1,39 @@
...
+  },
+];
+
 const MilestonesComponent: React.FC = () => {
-  return <div className="Milestones">Hello from MilestonesComponent</div>;
+  return (
+    <div className="Milestones">
+      <h1 className="Milestones__title">Milestones of 2020</h1>
+      <div className="Milestones__container">
+        {Object.entries(milestones).map(([_, milestone]) => (
```
ilopezro, line 28: Since you created an array of objects, there's actually no need to do `map(([_, milestone])...`. You can simply just do 
```suggestion
        {Object.entries(milestones).map(({title, text}) => (
```
or
```suggestion
        {Object.entries(milestones).map((milestone) => (
```
I personally like de-structuring the objects since it creates slightly cleaner code when you go ahead and use the components/properties on lines 29-32

src/views/Home/components/Milestones/index.view.tsx
ilopezro, line 32: If we destructure the object, these lines of code become much cleaner in a sense that instead of calling `milestone.text/title` we can simply have
```suggestion
          <div className="Milestones__box" key={title}>
            <h1 className="Milestones__box--top">{text}</h1>
            <p className="Milestones__box--bottom">{title}</p>
          </div>
```

src/views/Home/components/Milestones/milestones.scss
```diff
@@ -1,3 +1,59 @@
...
+    min-width: 200px;
+    text-align: center;
+  }
+
+  &__box--top {
+    @include typography.title-text;
+
+    padding: 30px 45px;
+    background: rgb(96, 73, 155);
+    text-shadow: 3px 3px #0085ff, -2px -2px rgba(255, 0, 0, 0.57);
```
ilopezro, line 38: I want to second Tim's comment above here as well. Although there isn't much repetition (just twice), we should put these in variables if we can. The hex & rgba values can go into a variable inside `variables.scss` and I could even go as far as saying that `3px 3px #0085ff, -2px -2px rgba(255, 0, 0, 0.57)` can be put into its own variable within this class as well!

src/views/Home/index.view.tsx
```diff
@@ -3,6 +3,7 @@ import HeroView from "views/Hero/index.view";
 import Button from "components/Button/index.view";
 import EmailSubscriptionForm from "components/EmailSubscription/index.view";
 import MLHBanner from "components/MLHBanner/index.view";
+import MilestonesComponent from "./components/Milestones/index.view";
```
ilopezro, line 6: No need to include `./` in here as part of your path since all paths are relative to `src` 

```suggestion
import MilestonesComponent from "components/Milestones/index.view";
```

src/views/Home/components/Milestones/index.view.tsx
w-gao: Awesome! This is much cleaner. 👍

src/views/Home/components/Milestones/milestones.scss
w-gao: I agree that we should use variables for these! Any suggestion for naming? Some of the colors are only relevant to a few components w/ some variations.

src/views/Home/index.view.tsx
w-gao: I'm keeping this because this is in the `components/` directory inside `views/Home`. 

Otherwise it'd be `import MilestonesComponent from "views/Home/components/Milestones/index.view";`

ilopezro: No you're totally right here! I just realized you're referencing the components within the `Home` view. This is my bad!

src/views/Home/components/Milestones/milestones.scss
ilopezro: name suggestions for the color is really up to you. I really try and eyeballing it and giving it a name and just writing comments in the variables about where it's being used.

src/views/Home/components/Milestones/milestones.scss
```diff
@@ -1,3 +1,81 @@
...
+
+    margin: 0;
+    padding: 10px 20px;
+    background: linear-gradient(180deg, var(--light-purple) 0%, var(--light-purple__gradient) 100%);
+    box-shadow: 0 0 35px -3px var(--heliotrope);
+    text-shadow: 1px 1px rgba(0, 0, 0, 0.15);
+    border-radius: 0 0 35px 35px;
+  }
+
+  &__pencil {
```
mocklee, line 58: I suggest adding a container for all these accent graphics with a `max-width` that is appropriate alongside the milestone box `container`.  here's what it looks like at 3MP:
Screenshot from 2020-12-04 15-18-16

src/views/Home/components/Milestones/milestones.scss
```diff
@@ -1,3 +1,81 @@
-/* stylelint-disable-line no-empty-source */
+@use "~stylesheets/typography";
+@use "~stylesheets/responsiveness";
 
-/* Delete these lines when you begin development */
+.Milestones {
+  overflow: auto;
+  margin: 0;
+  min-height: 375px;
+  background: url("~images/components/milestones/background.png") repeat-y, var(--dark-purple);
```
mocklee, line 8: let's not condense the component background into a single png. consider pulling out the grid & lens flare sphere SVGs from Figma & layering them with the accent graphics in the container mentioned in my other comment.

mocklee (commented):
great progress & mobile responsiveness! just had some comments on the background implementation & ultrawide responsiveness.

src/views/Home/components/Milestones/milestones.scss
w-gao: Good catch! I wrapped everything inside a container with a `max-width` instead, so we can position the graphics relative to the component. (Otherwise a container for the graphics would just have height 0 since all elements are absolutely positioned, and `bottom` wouldn't work). 

This should also be consistent with the `HeroView`. 

Let me know if that works.

src/views/Home/components/Milestones/milestones.scss
w-gao: I can't find a way to repeat a SVG vertically based on height without using the background property. Though, I replaced the png with svg in my newer commit. 

Let me know if there's a better way!

src/images/components/milestones/shell.svg
```diff
@@ -0,0 +1,134 @@
...
+    <path d="M359.186 381.964C372.418 394.625 380.73 405.3 378.698 407.64C376.665 409.981 364.997 403.233 350.544 391.92C337.289 379.258 328.978 368.572 330.998 366.243C333.019 363.914 344.778 370.627 359.186 381.964Z" fill="#1A1542"/>
+    <path d="M402.627 368.835C400.789 373.162 398.916 377.283 397.078 381.142C389.086 397.628 381.54 408.839 378.686 407.629C375.832 406.419 378.594 393.187 384.931 376.005C386.415 371.997 388.082 367.785 389.92 363.446C399.624 340.533 410.333 323.157 413.895 324.629C417.457 326.102 412.331 345.91 402.627 368.835Z" fill="#497B91"/>
+    <path d="M397.078 381.143C389.086 397.628 381.54 408.839 378.686 407.629C375.832 406.419 378.594 393.187 384.931 376.005C392.922 359.496 400.48 348.297 403.312 349.495C406.143 350.694 403.403 363.937 397.078 381.143Z" fill="#1A1542"/>
+    <defs>
+        <linearGradient id="gradient1" x1="135.196" y1="154.484" x2="261.305" y2="303.451" gradientUnits="userSpaceOnUse">
```
w-gao, line 42: Just a note, I changed these manually from `paintXXX_linear` to `gradientXXX` so the colors show up correctly in Chrome. Not sure how it works though.

Link to solution: https://stackoverflow.com/a/60249151

src/views/Home/components/Milestones/milestones.scss
```diff
@@ -1,3 +1,85 @@
...
-/* Delete these lines when you begin development */
+.Milestones {
+  margin: 0;
+  background: url("~images/components/milestones/background.svg") repeat-y, var(--dark-purple);
+  background-size: 100% auto;
+
+  &__container {
+    position: relative;
+    max-width: 1800px;
+    min-height: 400px;
```
tim-nguyen-cs, line 12: Would it be possible to change `max-width` to at least `2500px`? On an wide display, the pencil overlaps with the milestone boxes. We might also want to increase `min-height` to include all images at this screen resolution too.

tim-nguyen-cs (commented):
Nice work so far @w-gao ! It's looking great - my only suggest would be to see if we change either create more dynamic breakpoints for the images or find out how we can create some function to decrease the size of each image proportionally. I'm facing this problem too and it _might_ be resolvable with `flex-shrink` but I'll let you know how it goes!

src/views/Home/components/Milestones/milestones.scss
w-gao: How does this look? I changed `max-width` to `2000px` instead because I feel like `2500px` is too wide. I also added a few more breakpoints for the images but that might still need tweaking. Please let me know if `flex-shrink` works!

tim-nguyen-cs: For some reason, it still looks weird on my end, but I'll trust your intuition! 
image

Other than that, this looks good to me! Let's wait for @ilopezro or @mocklee's call!

src/views/Home/components/Milestones/milestones.scss
w-gao: @mocklee Continuing https://github.com/CruzHacks/cruzhacks-2021-website/pull/97#discussion_r539613066, `preserveAspectRatio="none"` doesn't quite work because it would stretch the grid. I'm using `preserveAspectRatio="xMinYMax slice"` for now so the bottom would stay consistent. The top would be cropped in some cases but that's the best that I can do. Please let me know if I'm missing anything.

mocklee: yea `xMinYMax slice` is a pretty clean solution, grid lines are indeed tricky.. 

this implementation of the background grid SVG is probably as good as it'll get without tossing out SVG altogether & drawing responsive squares with CSS (which we probably don't have time to do 😅️)

mocklee (approved):
we're good to stage from here! awesome work, things are starting to look pretty wicked

PR #92, opened 2020-12-02
Boilerplate background component layers

PriyaRajarathinam:
Problem
=======
As an engineer, I want a baseline background for cross-sectional components.

[Link to Asana Ticket](https://app.asana.com/0/1197738625416321/1199332791703155)

Solution
========

I implemented a Background component that creates an overarching background for the sections. There is a separate background for Sections 1-4 and a separate background for Sections 5-7. Sections 1-4 have the basic background with a blue wire frame and Sections 5-7 have a gradient background with a rainbow wire frame.

Change summary:
---------------
* Added background component and imported new images
* added background to the home component

Dev-Ops
=======
- [ ] `.env` & `.env.sample` 
- [ ] `webpack.config.js` 
- [ ] GCP Secret Manager
- [ ] Github Secrets 
- [ ] GithubActions Workflow

Steps to Verify:
----------------
1. Start the webserver to see the feature
2. It looks nice with only seven sections (4 in section 1 and 3 in section 2)

Screenshots (optional):
-----------------------


tim-nguyen-cs:
The static check is failing, because I think @ilopezro implemented branch naming conventions. I would suggest opening another pull request with the appropriate branch name.

src/components/Background/index.view.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { ReactComponent as Rainbow } from "images/rainbowpath.svg";
+interface BackgroundProps {
+  children?: React.ReactNode;
+}
+
+const Background: React.FC<BackgroundProps> = ({
+  children,
+}: BackgroundProps) => {
+  return (
+    <div>
```
tim-nguyen-cs, line 13: This just looks like an empty `div` container. If it doesn't do very much, I would probably remove it.

src/components/Background/index.view.tsx
```diff
@@ -0,0 +1,34 @@
...
+      <div className="Background">
+        <div className="Background__children">
+          {Array.isArray(children)
+            ? children.filter((child, index) => index < 4)
+            : children}
```
tim-nguyen-cs, line 18: A more efficient approach might be to require that `children` is an array, and then conditionally render if it exists instead of using a ternary operator.

ilopezro:
Yeah, we can go ahead and close this PR up! There's another one open in #93!

ilopezro:
We can go ahead and close this PR up since we've got another one open in #93!

PR #93, opened 2020-12-02
Feature/background component

PriyaRajarathinam:
Problem
=======
As an engineer, I want a baseline background for cross-sectional components.

[Link to Asana Ticket](https://app.asana.com/0/1197738625416321/1199332791703155)

Solution
========

I implemented a Background component that creates an overarching background for the sections. There is a separate background for Sections 1-4 and a separate background for Sections 5-7. Sections 1-4 have the basic background with a blue wire frame and Sections 5-7 have a gradient background with a rainbow wire frame.

Change summary:
---------------
* Added background component and imported new images
* added background to the home component

Steps to Verify:
----------------
1. Start the webserver to see the feature
2. It looks nice with only seven sections (4 in section 1 and 3 in section 2)

Screenshots (optional):
-----------------------


src/components/Background/index.view.tsx
```diff
@@ -0,0 +1,34 @@
+import React from "react";
```
ilopezro, line 1: You should take a look at the changes introduced in PR #71 when upgrading `typescript` to 4.1+. We don't need to import react anymore unless we need it. Typescript, in general, we import react the following way: 
```suggestion
import * as React from "react";
```

src/components/Background/index.view.tsx
```diff
@@ -0,0 +1,34 @@
...
+const Background: React.FC<BackgroundProps> = ({
+  children,
+}: BackgroundProps) => {
+  return (
+    <div>
+      <div className="Background">
+        <div className="Background__children">
+          {Array.isArray(children)
+            ? children.filter((child, index) => index < 4)
+            : children}
```
ilopezro, line 18: There's no need to check to see if `children` prop is an array. We can simply just drop the `children` props in here

src/images/wire.svg
```diff
@@ -0,0 +1,12 @@
...
+<defs>
+<linearGradient id="paint0_linear" x1="1017.01" y1="0.138689" x2="357.536" y2="3536.37" gradientUnits="userSpaceOnUse">
+<stop offset="0.0165518" stop-color="#4366C0" stop-opacity="0"/>
+<stop offset="0.10046" stop-color="#4366C0" stop-opacity="0.3"/>
+<stop offset="0.244792" stop-color="#4366C0" stop-opacity="0.3"/>
+<stop offset="0.837272" stop-color="#4366C0" stop-opacity="0.3"/>
+<stop offset="0.963542" stop-color="#4366C0"/>
+</linearGradient>
+</defs>
+</svg>
```
ilopezro, line 12: Super nit picky, but could we add an end-line here?

src/stylesheets/_variables.scss
```diff
@@ -4,13 +4,16 @@
...
+  --darkmode: #1a1542;
+  --blue: #4366c0;
 
   /* variations */
   --wedgewood__dark: #{darken(#4f728e, 7%)};
 
   /* These are for gradient */
   --wedgewood__gradient: #486988;
   --waterleaf: #a9ebdf;
+  --blue_gradient: #344790;
```
ilopezro, line 16: In order to be more in line with the BEM naming conventions, we should have this be `blue__gradient`(with the double underscore) or `blueGradient`

src/views/Hero/Hero.scss
```diff
@@ -7,6 +7,7 @@
   min-height: 100%;
 
   &__homeView {
+    /*
```
ilopezro, line 10: I'm wondering if we could just delete all these comments. If we no longer need these lines of code, let's just get rid of them!

src/components/Background/index.view.tsx
PriyaRajarathinam: If I simply drop the children props how would I section the pages?

ilopezro: I think "drop" would be the wrong word here. What I was trying to say was that we can simply have this block be: 
```suggestion
          {children}
```

src/components/Background/index.view.tsx
```diff
@@ -0,0 +1,22 @@
...
+const Background: React.FC<BackgroundProps> = ({
+  children,
+}: BackgroundProps) => {
+  return (
+    <div className="Background">
+      <div className="Background__gradient"></div>
+      <div className="Background__children">{children}</div>
+      <Wire className="Background__wire"></Wire>
+      <Rainbow className="Background__wire--variant"></Rainbow>
+    </div>
```
ilopezro, line 18: 🎉

ilopezro (commented):
This looks really good to me! I just have a super quick question: is this the intended view right now? I see that when I resize I can see the rest of the background. My guess is that as we keep adding components, the background will slowly appear where it's supposed to be

Screen Shot 2020-12-03 at 12 56 22 AM

src/components/Background/index.view.tsx
PriyaRajarathinam: Yup! It's meant to work with 7 sections.

w-gao: This looks amazing with more sections added in!

mocklee (commented):
cool progress so far! let's flush out further details with Calum tomorrow during the meeting

src/components/Background/Background.scss
```diff
@@ -0,0 +1,40 @@
+@use "~stylesheets/typography";
+
+.Background {
```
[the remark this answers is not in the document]
tim-nguyen-cs: Could we even do `7000px` for each of the supposed 7 sections?

src/components/Background/Background.scss
```diff
@@ -0,0 +1,40 @@
...
+    z-index: -2;
+    position: absolute;
+    background: linear-gradient(
+      180deg,
+      var(--blue) 0%,
+      var(--blue__gradient) 100%
+    );
+    width: 100%;
+    height: calc(100% / 7 * 3);
+    bottom: 0;
```
tim-nguyen-cs, line 38: Do you think we could rearrange the SCSS to replicate the structure of the HTML? It'll just make things easier to understand.

src/components/Background/Background.scss
```diff
@@ -0,0 +1,40 @@
...
+
+    &--variant {
+      position: absolute;
+      width: 100%;
+      bottom: 0;
+    }
+  }
+
+  &__children {
+    z-index: 999;
```
tim-nguyen-cs, line 24: Can we also change to to something more realistic? Like 0 or 1? My components have z-indices too, but I'm not sure how that might clash.

tim-nguyen-cs (commented):
Everything looks good to me! I love the absolute/relative positioning magin - great job @PriyaRajarathinam !

src/components/Background/Background.scss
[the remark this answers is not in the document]
mocklee: hopefully we won't need to define a height if the sections size up well against the background

mocklee (approved):
I think the background is ready for staging! let's get this into `main` & go from there..

PR #94, opened 2020-12-05
Mission section

tim-nguyen-cs:
Problem
=======
As a prospective hacker,
I would like to read CruzHacks' mission statement
To understand more about the organization and event.

[Link to Asana Ticket](https://app.asana.com/1234)

Solution
========
To implement the `Mission` section of the homepage, flexbox-based positioning, (negative) margins, and other CSS styling techniques were used to replicate the provided Figma design. Text color variables and other box variations will be replaced after pre-existing pull requests are merged in. In the `Home` component, `MissionProps` corresponding to dictionary objects that represent each title and body text are then passed into the corresponding component as input parameters.

Change summary:
---------------
* finished declaring + styling mission statements f85dff0

Steps to Verify:
----------------
1. Checkout branch using `git checkout feature/mission`
3. Verify that mission section is designed and responds correctly.


src/views/Home/index.view.tsx
```diff
@@ -51,20 +21,10 @@ const HomepageView: React.FC = () => {
...
-          <div className="Homepage__buttonContainer">
-            {Object.entries(buttonProps).map(([_, buttonProp]) => (
-              <Button
-                key={buttonProp.text}
-                className="Homepage__button"
-                {...buttonProp}
-              />
-            ))}
-          </div>
-        </HeroView>
```
ilopezro, line 67: Not entirely sure why we got rid of this -- this PR/branch deals with the `Mission` component only, so let's put this back!

tim-nguyen-cs: Ahh, another Freudian slip! I was thinking that this section was going to be replaced instead of reprovisioned. Nice catch!

src/views/Home/components/Mission/index.view.tsx
```diff
@@ -1,8 +1,46 @@
...
+}
+
+const MissionComponent: React.FC<MissionProps> = ({
+  about_text,
+  mission_text,
+}: MissionProps) => {
+  return (
+    <div className="Mission">
+      <div className="Mission--text">
+        <div className="Mission__section">
```
w-gao, line 19: I think we should have the same styling for both sections to prevent something like this:

w-gao (commented):
This looks great! See below for comments.

src/views/Home/components/Mission/index.view.tsx
tim-nguyen-cs: The sections are slightly different because of the gradient highlight, but I fixed it with `max-width`!

src/views/Home/components/Mission/mission.scss
```diff
@@ -1,3 +1,115 @@
...
+
+  /* Image Layer */
+  &--images {
+    display: flex;
+    flex-direction: column;
+    width: 100%;
+
+    position: absolute;
+    top: 0;
+    left: 0;
```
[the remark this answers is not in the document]
tim-nguyen-cs: That sounds great! Thanks for the help - looks much better now!

src/views/Home/components/Mission/mission.scss
```diff
@@ -1,3 +1,108 @@
...
+.Mission {
+  position: relative;
+
+  /* Text Layer */
+  &--text {
+    display: flex;
+    flex-direction: column;
+    background: url("~images/mission-background.png");
+    background-size: 100%;
+    padding: 10% 22%;
```
w-gao, line 14:
```    
    @include responsiveness.respond(phone) {
      padding: 10px 20px;
    }
```

src/views/Home/components/Mission/index.view.tsx
w-gao: Ahh, that's what I thought. I was hoping that we can use the same structure for both sections so we can put the mission statements inside a list and `map` them instead (where only the first section would include the `Mission__highlight` div). But I'm totally fine with what you have now since we only have two sections. 

However, I could be wrong but I don't think the `max-width` in `&--subsection` is applied anywhere? I think a temporary fix is to add a fixed padding size in `&--text` for smaller screens so the entire thing can fit in 320px. The real fix might has something to deal with the flex display which I'm not entirely sure.

w-gao (commented):
Great work so far @tim-nguyen-cs! Other than the responsiveness suggestion, I think this looks pretty good to me.

src/views/Home/components/Mission/index.view.tsx
w-gao: Ahh, that's what I thought. I was hoping that we can use the same structure for both sections so we can put the mission statements inside a list and `map` them instead (where only the first section would include the `Mission__highlight` div). But I'm totally fine with what you have now since we only have two sections.

However, I could be wrong but I don't think the `max-width` in `&--subsection` is applied anywhere? I think a temporary fix is to add a fixed padding size in `&--text` for smaller screens so the entire thing can fit in 320px. The real fix might has something to deal with the flex display which I'm not entirely sure.

tim-nguyen-cs: Ahh, sorry about that! That's actually cause I called it `sub-container` instead of `subsection`. The most recent push should solve it, but let me know if you see anything else! Thank you for the help!

src/views/Home/components/Mission/mission.scss
tim-nguyen-cs: Personally, I think the existing iteration is better, because it doesn't overlap with the images:
image

I would love to hear your feedback though - we can scale the `title-text` appropriately later.

w-gao: Generally, I would think it is ok to drop graphics on a limited size screen so more space can be used to display information, but this looks fine as well. Thanks for making the adjustments!

tim-nguyen-cs: No, yeah - that makes sense! @ilopezro & @mocklee, any opinions?

mocklee: apologies for not getting on this sooner.. I'm not opposed to leaving it as-is, the 3d effect works pretty well with how the font is rendering!

src/views/Home/components/Mission/mission.scss
```diff
@@ -1,3 +1,108 @@
...
 
-/* Delete these lines when you begin development */
+.Mission {
+  position: relative;
+
+  /* Text Layer */
+  &--text {
+    display: flex;
+    flex-direction: column;
+    background: url("~images/mission-background.png");
```
mocklee, line 12: check out the changes made [here](https://github.com/CruzHacks/cruzhacks-2021-website/pull/91#discussion_r536441553) to remove the hardcoded background color & add transparency while keeping the grid svg— this will let the page-level background float visibly underneath
Screenshot from 2020-12-19 13-44-22

src/views/Home/components/Mission/mission.scss
```diff
@@ -1,3 +1,108 @@
...
+    width: 18%;
+    left: 80%;
+    top: 32%;
+  }
+
+  &__flower {
+    position: absolute;
+    width: 28%;
+    left: 5%;
+    top: 43%;
```
mocklee, line 106: consider throwing in a relative unit here like `vh` using the `calc` function! just using it on `top` puts the flower in places with better contrast across mobile & desktop.
Screenshot from 2020-12-19 13-29-53

src/views/Home/components/Mission/mission.scss
```diff
@@ -1,3 +1,108 @@
...
+@use "~stylesheets/responsiveness";
+@use "~stylesheets/typography";
+@use "~stylesheets/variables";
 
-/* Delete these lines when you begin development */
+.Mission {
+  position: relative;
+
+  /* Text Layer */
+  &--text {
```
mocklee, line 9: we should have a `max-width` to constrain how wide the paragraph renders at larger screens.. I noticed adding a `max-width` here doesn't apply to the second mission section in the same way, so the implementation is worth revisiting.. the DOM structure is worth applying here to keep the widths uniform across sections 
Screenshot from 2020-12-19 13-37-18

mocklee (changes requested):
we're super-close with this component— just need a revision of the background & text container widths here. remember that containers need bounds to respond across resolutions

src/views/Home/components/Mission/mission.scss
mocklee: also consider revising these columns so they're wider & friendlier to read at this breakpoint!

src/views/Home/components/Mission/mission.scss
tim-nguyen-cs: I've added breakpoints to decrease spacing and slightly increase the container widths (even though it might be off-centered now). Let me know how it looks!

src/views/Home/components/Mission/mission.scss
tim-nguyen-cs: I went ahead an used `bottom` instead - let me know how it looks!

src/views/Home/components/Mission/mission.scss
tim-nguyen-cs: I'm not too sure why you're seeing it like that - this is how my view looks like on a large display. Can you check on it again?

image

src/views/Home/components/Mission/mission.scss
tim-nguyen-cs: Thanks for letting me know about that - it looks so much better!

src/views/Home/components/Mission/mission.scss
mocklee: this is the case I'm referring to: the text containers aren't bounded, and the paragraphs turn into one long sentence
Screenshot from 2020-12-23 16-36-04

mocklee: this is what the hero-in-progress & milestones looks like at the same resolution since those components have width constraints on the containers
Screenshot from 2020-12-23 16-42-44

src/views/Home/components/Mission/mission.scss
w-gao: This looks great, thanks for making the adjustments 🙂.

src/views/Home/components/Mission/mission.scss
tim-nguyen-cs: Gotcha, I think I resolved it - let me know how it looks!

mocklee (approved):
let's get this to staging!

PR #95, opened 2020-12-06
navigation bar

tim-nguyen-cs:
Problem
=======
As a prospective hacker,
I would like to view resources in a navigation bar
To more easily navigate the hackathon homepage.

[Link to Asana Ticket](https://app.asana.com/0/1197738625416321/1198219556358101/f)

Solution
========
The navigation bar was implemented previously, but a few things were updated to make this integration more closely represent the provided design. First, the default button styling was updated to appear sleeker and more modern. Then, I changed `buttonProps` to include an enumerated `ButtonType` property to separate primary and secondary buttons (also removing the previous button map from the `Home` page to reduce redundancy; primary buttons are left-aligned and secondary buttons are right-aligned. 

Because a mobile version hasn't been implemented yet, each button just wraps for now, but feedback would be greatly appreciated on how we can scale button sizes more appropriately.

Change summary:
---------------
* first iteration of navigation bar 069b252

Steps to Verify:
----------------
1. Checkout branch using `git checkout feature/navbar`
3. Verify that the navigation bar section is designed correctly and responds accurately.


mocklee (changes requested):
We need to work on responsiveness a bit more with the navbar— consensus from #design would be nice. Maybe a hamburger menu would be useful here.. some people find that button annoying but it would definitely make things cleaner & more tappable:

Screenshot from 2020-12-11 23-22-21

tim-nguyen-cs:
Yep! I was considering a hamburger button too but wanted to have the iteration up as quickly as possible. I can work on that in the next few days if it's pretty straightforward.

mocklee (approved):
Great work implementing a hamburger menu! 

I do have some nitpicks with how links are rendering right before the `tab-land` breakpoint— it may be worth it to always render a hamburger so we can hide secondary links like `contact us` and `code of conduct` within, while collapsing the other primary CTA links at the breakpoint.

Screenshot from 2020-12-31 11-09-25

tim-nguyen-cs:
I updated the menu button to be persistent based on screen sizes and used another library to help set responsiveness in React. I think it's a little jankey though, so let me know if there's an easier way!

src/components/Navbar/index.view.tsx
```diff
@@ -8,14 +16,96 @@ interface NavbarProps {
...
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
+
+  const useStyles = makeStyles({
+    buttonWrapper: {
+      margin: "0 calc(5% + var(--small-spacer)) 0 auto",
```
mocklee, line 31: The margin here isn't responsive enough to the viewport width at tablet breakpoints; there should be a bit more space between the banner & hamburger button, especially since both are clickable links
Screenshot from 2021-01-02 17-21-10

mocklee (changes requested):
> I updated the menu button to be persistent based on screen sizes and used another library to help set responsiveness in React. I think it's a little jankey though, so let me know if there's an easier way!

Looks like you took a bit more of an imperative approach to styling & composition by leveraging Javascript— unfortunately that confines CSS declarations into Javascript strings, but it's just a different paradigm that has its ups & downs. Just have one comment regarding the button margin on mobile, but this looks like a solid implementation to me. (I do see some weird behavior with the scrollbar when opening the menu, but that seems to just be a quirk from the UI library)

mocklee (approved):

PR #97, opened 2020-12-07
Prizes component.

w-gao:
Problem
=======
As a hacker, 
I want to see the available prices
So I can get excited about them!

[Link to Asana Ticket](https://app.asana.com/0/1197738625416321/1199333062665014)

Solution
========
The layout is similar to the Milestones component, expect this one doesn't have any images/graphics associated with it. ~~I had to copy over some of the mixins/variables from #91, but they will be removed after that PR is pulled in.~~

In the figma design, the first half of the Prizes section belongs to the Milestones section. I had to use separate backgrounds for the two sections to not deal with negative margins. The integrated design looks like this: 

I don't see a clear way to share a background with another component without messing up the spacing somehow, but feedback / suggestions would be greatly appreciated. 

Change summary:
---------------
* Prizes component. https://github.com/CruzHacks/cruzhacks-2021-website/commit/24e95233070d4b2a6a37529e67ae9fe9e5472be8

Steps to Verify:
----------------
1. Checkout branch and run `yarn run start` to start a local web server.
2. Verify the Prizes section is present and responsive.


src/stylesheets/_typography.scss
```diff
@@ -18,6 +19,28 @@ $font-family-italic: "Montserrat Italic", sans-serif;
...
+
+///////////
+
+/* smaller than title-text but bigger than description */
+@mixin title-text-secondary {
+  margin: 0;
+  color: var(--white);
+  font-family: $font-family-italic;
+  font-style: italic;
+  font-weight: normal;
```
tim-nguyen-cs, line 39: Instead of duplicating code here, can we extend from the mixin already defined? We might also be able to use our `--` notation, but I'm also not sure if that's correct again.

tim-nguyen-cs:
Nice job so far, @w-gao! My recommendation to overlap fields would to probably use some negative margin if you haven't already to move this section up. Would it also be possible to adjust the padding/dimensions of each prize box to hold a square (instead of vertically rectangular) shape?

Right now, it looks like this on my screen, so I'm hoping it'll be more proportioned when the rest of the sections are in:
image

src/stylesheets/_typography.scss
mocklee: the two dashes `--` is a [CSS naming convention for variables](https://www.w3schools.com/css/css3_variables.asp#main:~:text=Note%3A%20The%20variable%20name%20must%20begin%20with%20two%20dashes%20(%2D%2D)%20and%20it%20is%20case%20sensitive!). propose any changes you find to be worthwhile!

src/views/Home/components/Prizes/prizes.scss
```diff
@@ -1,3 +1,70 @@
-/* stylelint-disable-line no-empty-source */
+@use "~stylesheets/typography";
+@use "~stylesheets/responsiveness";
 
-/* Delete these lines when you begin development */
+.Prizes {
+  margin: 0;
+  background: url("~images/components/prizes/background.svg") repeat-y;
```
mocklee, line 6: Again, I'd advise against confining the background to a single CSS prop; it looks like the image is being cropped. A CSS class would lend enough granularity to scale the entire image! The background layers have a bit more complexity than a single graph; I think a separate background container will really help here.

implementation:
Screenshot from 2020-12-09 09-22-49

design:
Screenshot from 2020-12-09 09-23-13

mocklee: Also we can include the glowing shadow effect using CSS! https://www.w3schools.com/css/css3_shadows.asp

mocklee (changes requested):
Awesome, this component is definitely most of the way there! Just had some comments regarding the background & shadow effects.

src/views/Home/components/Prizes/prizes.scss
w-gao: @mocklee A separate background container would make sense so I will update #91 and this PR. However, I'm still unsure how to repeat the SVG without using the `background-repeat` prop. This was needed so the background can cover the entire section on small screens.

mocklee: instead of repeating the grid SVG, its height & width can be set proportional to a container. if it doesn't size properly on smaller screens, consider jumping into the SVG & defining `preserveAspectRatio` to lend a bit more flexibility with resizing: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/preserveAspectRatio

src/views/Home/components/Prizes/prizes.scss
```diff
@@ -1,3 +1,87 @@
...
+
+  &__boxContainer {
+    display: flex;
+    flex-wrap: wrap;
+    justify-content: center;
+    max-width: 1500px;
+    margin: var(--medium-spacer) auto;
+  }
+
+  &__box {
```
w-gao, line 37: @tim-nguyen-cs Let me know how this looks: 



I'm not enforcing a height here because it has to be rectangular on smaller displays to fit everything, but I think it is squarish on most of the screen sizes now.

tim-nguyen-cs: Looks great!

mocklee (approved):
Great work tying this in with the milestones component. I do suggest using the mobile breakpoints to keep the grid somewhat present & aligned with the grid in the milestones component instead of having the grid disappear. it's a small aesthetic issue on mobile we may not have time to address, but worth noting!

src/views/Home/components/Prizes/index.view.tsx
```diff
@@ -1,8 +1,72 @@
 import React from "react";
```
ilopezro, line 1: #71 allows us to remove this statement completely if we do not use react within a file; however, since we use it here we can simply do: `import * as React from "react";`. With newer versions of react and typescript coming out, if react is not used within a file, you don't have to import react. It's pretty neat stuff!

PR #107, opened 2020-12-10
Background responsiveness & behavior

mocklee:
Problem
=======
The background component is substantiated thanks to @PriyaRajarathinam! Just a few responsiveness details to address here.

The hero component will be scaffolded on this branch as well, since it's partially dependent on the wire sticking at the top of the page.

Solution
========
This patch sets the background wire `preserveAspectRatio` to `none` & pins the wire at the top of the page, so it dynamically resizes in an `absolute` position.

Screenshot from 2020-12-09 23-21-45

Screenshot from 2020-12-09 23-22-29


src/images/rainbowpath.svg
```diff
@@ -1,151 +1,139 @@
...
...6C2603.42 1320.36 2603.51 1320.55 2603.58 1320.75C2603.65 1320.89 2603.71 1321.03 2603.76 1321.18V1321.18C2603.83 1321.33 2603.89 1321.49 2603.95 1321.65C2604.34 1322.57 2604.72 1323.49 2605.1 1324.41C2605.77 1326.05 2606.44 1327.69 2607.1 1329.34C2607.51 1330.34 2607.91 1331.34 2608.3 1332.42C2608.63 1333.28 2608.96 1334.13 2609.3 1334.99C2609.47 1335.43 2609.64 1335.87 2609.8 1336.32C2610.09 1337.08 2610.37 1337.86 2610.65 1338.63C2611.41 1340.71 2612.15 1342.8 2612.87 1344.9C2613.23 1345.9 2613.59 1347.01 2613.94 1348.06C2614.19 1348.79 2614.43 1349.53 2614.67 1350.26C2614.67 1350.36 2614.74 1350.45 2614.77 1350.55L2615.16 1351.76L2615.22 1351.96L2616 1354.42C2616.62 1356.36 2617.21 1358.32 2617.78 1360.28C2618.05 1361.18 2618.31 1362.09 2618.57 1362.99C2618.99 1364.43 2619.39 1365.87 2619.78 1367.32C2620.05 1368.32 2620.32 1369.32 2620.58 1370.32C2620.76 1370.98 2620.94 1371.65 2621.1 1372.32C2621.26 1372.99 2621.37 1373.37 2621.5 1373.89C2621.72 1374.74 2621.93 1375.6 2622.14 1376.46L2622.58 1378.3C2622.86 1379.45 2623.13 1380.6 2623.38 1381.75C2623.58 1382.59 2623.76 1383.43 2623.94 1384.27C2624.12 1385.11 2624.32 1385.97 2624.49 1386.83C2624.71 1387.84 2624.92 1388.84 2625.1
```
w-gao, line 127: One side note: It seems like these linear gradient and filter ids would conflict between SVG files so it might be useful if we make them unique. 

I've been manually renaming them (e.g.: https://github.com/CruzHacks/cruzhacks-2021-website/pull/91/files#diff-5a5df358dc51c0b9b5722d062a56e0c8f8bb1dff0eb1cc4efd4d70b4b91f9cfcR8) but I think it is also possible to configure something to automatically attach a random id, though that may require more work.

w-gao (approved):
This looks good to me! See below for a semi-related comment.

src/images/rainbowpath.svg
mocklee: thanks for catching that @w-gao, this [NPM package](https://www.npmjs.com/package/react-svg-unique-id) gives us a neat wrapper for SVGs to dynamically create unique IDs. I'll update the branch with that drop-in component & hopefully it should resolve things.

@CruzHacks/2021-engineers we'll need to apply something like this to prevent ID collisions on SVG elements across everyone's components!

PR #109, opened 2020-12-12
Feature/schedule component

PriyaRajarathinam:
Problem
=======
As a hacker, 
I want to see a rough schedule of the event 
So I can decide if it fits with my schedule 

[Link to Asana Ticket](https://app.asana.com/0/1197738625416321/1199333062665016)

Solution
========
Create a view that has the schedule for each day.

Change summary:
---------------
* Created rectangles
* Added JSON content
* added events to the schedule

Steps to Verify:
----------------
1.  yarn start
2. view schedule component

Screenshots (optional):
-----------------------
Screenshot
https://puu.sh/GWVU7/c2da09838d.png


src/views/Home/components/Schedule/index.view.tsx
```diff
@@ -1,8 +1,101 @@
...
+      prevActive.classList.remove("Schedule__tab--variant");
+      prevActive.style.zIndex = "-3";
+    }
+    const currActive = document.getElementById(`day${day}`);
+    if (currActive) {
+      currActive.classList.add("Schedule__tab--variant");
+      currActive.style.zIndex = "0";
+      setSelectedDay(day - 1);
+    }
+  };
```
ilopezro, line 43: If this is what we are using to style and change class names, take a look at the [`classnames`](https://www.npmjs.com/package/classnames) package and how it's used here: 
https://github.com/CruzHacks/cruzhacks-2021-website/blob/482cd786794a84bc5c79d8824a45ef87ff7c2749/src/views/Hero/index.view.tsx#L30-L32

In turn, having this logic here would allow you to do all the styling in the SCSS regardless of the conditionals in the TS code.

src/stylesheets/_typography.scss
```diff
@@ -19,6 +20,33 @@ $font-family-italic: "Montserrat Italic", sans-serif;
   }
 }
 
+@mixin scheduleheader-text {
```
tim-nguyen-cs, line 23: Because this text styling will be used in multiple places, can we use a more generic name and add color variations within the SCSS file itself? This will help reduce how repetitive our styles can get.

src/views/Home/components/Schedule/index.view.tsx
```diff
@@ -1,8 +1,135 @@
...
+import { useState } from "react";
+import * as React from "react";
 import "./schedule.scss";
+import cx from "classnames";
 
+const scheduleItems = [
+  {
+    date: "Friday, January 15, 2021",
+    events: [
+      { event: "Event Begins", time: "5:00 pm" },
```
tim-nguyen-cs, line 11: Can we change `event` to `name` to avoid having to call something like `event.event`?

src/views/Home/components/Schedule/index.view.tsx
```diff
@@ -1,8 +1,135 @@
...
+              "Schedule__dayThree"
+            )}
+          >
+            Day 3
+          </div>
+        </div>
+        <div className="Schedule__pinkRect"></div>
+        <div className="Schedule__whiteRect--variant"></div>
+      </div>
+      <img
```
w-gao, line 96: If possible, I think it would be better to import the SVGs as components.

src/images/schedule/grid.svg
```diff
@@ -0,0 +1,15 @@
+<svg preserveAspectRatio="none" viewBox="1170 -350 2640 1790" fill="none" xmlns="http://www.w3.org/2000/svg">
```
w-gao, line 1: Consider using `preserveAspectRatio="xMinYMax slice"`. I had the issue where the grid would stretch on mobile in my component as well.

w-gao (commented):
Great work! This looks super good but below are just a few more suggestions.

src/views/Home/components/Schedule/index.view.tsx
PriyaRajarathinam: Converting some of them to components seems to mess them up : ( 
image

src/stylesheets/_fonts.scss
```diff
@@ -39,3 +39,59 @@
...
+  font-family: "Source Code Pro";
+  font-style: normal;
+  font-weight: 400;
+  src: url("../fonts/source-code-pro-v13-latin-regular.eot"); /* IE9 Compat Modes */
+  src: local(""),
+    url("../fonts/source-code-pro-v13-latin-regular.eot?#iefix") format("embedded-opentype"), /* IE6-IE8 */
+    url("../fonts/source-code-pro-v13-latin-regular.woff2") format("woff2"), /* Super Modern Browsers */
+    url("../fonts/source-code-pro-v13-latin-regular.woff") format("woff"), /* Modern Browsers */
+    url("../fonts/source-code-pro-v13-latin-regular.ttf") format("truetype"), /* Safari, Android, iOS */
+    url("../fonts/source-code-pro-v13-latin-regular.svg#SourceCodePro") format("svg"); /* Legacy iOS */
```
ilopezro, line 54: All these imports can be done from `~fonts/...`. The power and magic of having our root at `src`!

src/views/Home/index.view.tsx
```diff
@@ -3,6 +3,7 @@ import Hero from "views/Home/components/Hero/index.view";
 import Button from "components/Button/index.view";
 import EmailSubscriptionForm from "components/EmailSubscription/index.view";
 import MLHBanner from "components/MLHBanner/index.view";
+import Schedule from "views/Home/components/Schedule/index.view";
```
ilopezro, line 6: We can do the import here from `./components/Schedule/index.view;` instead of referencing back to `views`

src/views/Home/components/Schedule/index.view.tsx
w-gao: Ahh, I see.

ilopezro (approved):
🚢

PR #122, opened 2020-12-19
Hero component 

mocklee:
Problem
=======
As a prospective attendee,
I want to immediately pick up a vibe of what CruzHacks will be like this year upon landing on the website.

[Link to Asana Ticket](https://app.asana.com/0/1197738625416321/1199533585396245)

Solution
========
Spooling up the finalized hero described in the mock [here](https://www.figma.com/file/9vYrREdbNaxL3bLlyHnR7m/CruzHacks-2021-Website?node-id=772%3A764)

Screenshots (optional):
-----------------------
Screenshot from 2020-12-24 21-55-17


src/views/Home/components/Hero/index.view.tsx
```diff
@@ -37,6 +42,12 @@ const Hero: React.FC<HeroProps> = ({
   return (
     <>
       <div className="Hero__container">
+        <Grid className="Hero__grid" />
+        <LargeFlare className="Hero__largeFlare" />
+        <MediumFlare className="Hero__mediumFlare" />
+        <SmallFlare className="Hero__smallFlare" />
+        <Computer className="Hero__computer" />
+        <Mouse className="Hero__mouse" />
```
tim-nguyen-cs, line 50: I see what you did here by abstracting the three different lens flares; however, do you think we should just consolidate them apart of some `Background.svg` to be consistent with the `Milestones` and `Mission` component?

Is it better that we've abstracted them? Do they scale the same?

src/views/Home/components/Hero/Hero.scss
```diff
@@ -6,13 +6,36 @@
   overflow: hidden;
   min-height: 100%;
 
+  @mixin flare {
+    position: absolute;
+    z-index: -3;
+    width: 110vw;
+    height: calc(100%);
```
tim-nguyen-cs, line 13: What's the difference between `calc(100%)` and `100%`?

src/views/Home/components/Hero/Hero.scss
```diff
@@ -25,8 +48,107 @@
...
+      left: 0;
+      max-width: calc(20vw + 15vh);
+      min-height: 90%;
+      max-height: 100%;
+    }
+
+    @include responsiveness.respond(smallest) {
+      visibility: hidden;
+    }
+  }
```
tim-nguyen-cs, line 72: Reference my previous comment about simiplying this structure with a single `*.svg` file.

src/views/Home/components/Hero/Hero.scss
```diff
@@ -25,8 +48,107 @@
...
+
+  &__computer {
+    position: absolute;
+    z-index: -1;
+    top: calc(-30em - 4vh);
+    width: 100%;
```
tim-nguyen-cs, line 78: What's the purpose of this?

src/views/Home/components/Hero/Hero.scss
```diff
@@ -6,13 +6,36 @@
...
   &__container {
     display: flex;
+    position: relative;
     flex-wrap: nowrap;
-    margin: auto;
-    max-width: 1800px;
+    margin: 0;
+    width: 100%;
     padding-top: calc(65px + 10vh);
+    padding-bottom: calc(65px + 10vh);
```
tim-nguyen-cs, line 35: Maybe decrease this so there's not as much of a noticeable gap between sections.

tim-nguyen-cs:
@mocklee, this component looks insanely good! I am excited to finally see our mock-up come together!

I made some comments about slight design adjustments and structure pointers, but this looks really great overall!

src/views/Home/components/Hero/index.view.tsx
mocklee: good question, I decided against using a general background SVG & broke out the background elements to add more dynamics to the responsiveness for better UX. we could even animate these later on by applying CSS classes to each individual flare "component" using [react-on-visible ](https://www.npmjs.com/package/react-on-visible)

src/views/Home/components/Hero/Hero.scss
mocklee: yea this is a bit of dead code, I'll clean this up

src/views/Home/components/Hero/Hero.scss
mocklee: responsiveness is more flexible using this stylesheet structure! check out how it responds across ultrawide to mobile

src/views/Home/components/Hero/Hero.scss
mocklee: @tim-nguyen-cs since this component is the hero, I tried to have the element occupy most of the viewport. our past websites had the hero occupy the entire viewport, but I haven't integrated that for this website, given this year's constraints on time spent designing & boilerplating

src/views/Home/components/Hero/Hero.scss
mocklee: @tim-nguyen-cs not quite sure which line you're referring to here

src/views/Home/components/Hero/index.view.tsx
tim-nguyen-cs: Ohh, that makes sense! If this is the case, we might have to refactor the other few components that we finished, but it sounds  great!

src/views/Home/components/Hero/Hero.scss
tim-nguyen-cs: Acknowledged!

src/views/Home/components/Hero/Hero.scss
tim-nguyen-cs: Oh, my bad! I was referring to the `max-width: 100%` and the `width: 100%`.

src/views/Home/components/Hero/Hero.scss
tim-nguyen-cs: Ohh, okay ~ that makes sense! In that case, we'd definitely need something to hint that there's more content "under the fold," but that also sounds like a good plan, aha.

PriyaRajarathinam (commented):
image
Should we have a consensus on what the grid should look like in smaller viewports? Currently, in the hero component, it seems to shrink and in the milestone component it expands

tim-nguyen-cs:
@PriyaRajarathinam - Well, the screenshot above isn't really accurate, because the `Mission` section should be squeezed in between that (and the grid on `Mission` is right-aligned).

What I think is a better question is to see if we can fix the different grid sizes. The `Milestones` grid is really small and the `Hero` grid is much less dense.

ilopezro:
Just a quick comment, again about the commit histories. This is a good example of why we want to keep them separate from one another. You've added a `package.lock` to this PR, which does not belong. Problem is, since you have the same commit history in #139 then it's double the work you have to do in order to clean up the PR's 😬

mocklee:
> @PriyaRajarathinam - Well, the screenshot above isn't really accurate, because the `Mission` section should be squeezed in between that (and the grid on `Mission` is right-aligned).
> 
> What I think is a better question is to see if we can fix the different grid sizes. The `Milestones` grid is really small and the `Hero` grid is much less dense.

There's definitely room to make the grid in `Mission` more responsive with mobile breakpoints; it's worth keeping continuity with the design. I also agree that we should adjust the grid density in `Milestones`.. @w-gao do you have any time to sync up the grid sizing in your components with hero & mission?

src/views/Home/components/Hero/Hero.scss
mocklee: ah that was a bit of a hacky workaround with responsiveness, I've since updated the `max-width` line with a different unit.

src/views/Home/index.view.tsx
```diff
@@ -1,12 +1,29 @@
...
+import ChooseSantaCruz from "images/sponsors/chooseSantaCruz.svg";
+import Comcast from "images/sponsors/comcast.svg";
+import eBay from "images/sponsors/ebay.svg";
+import MLH from "images/sponsors/mlh.svg";
+import Pinnacle from "images/sponsors/pinnacle.svg";
+import SouthSwell from "images/sponsors/southSwell.svg";
+import Stickermule from "images/sponsors/stickermule.svg";
+import Stratovan from "images/sponsors/stratovan.svg";
+import Wolfram from "images/sponsors/wolfram.svg";
+import FifteenSeventeen from "images/sponsors/1517.svg";
```
ilopezro, line 26: I think we can move this inside the `Sponsors` component

src/views/Home/index.view.tsx
```diff
@@ -78,33 +95,118 @@ const mission_props = {
   },
 };
 
+const sponsors = [
```
ilopezro, line 98: This should follow

src/views/Home/index.view.tsx
mocklee: I had the impression that the home index file would contain the props for our components since it's the view controller, and this already seems to be how data is serialized in our app. [I'd rather implement according to this established pattern](https://blog.andrewray.me/the-reactjs-controller-view-pattern/). 

If the increasing file size is your worry, we can modularize the props into JSON so data can be cleanly passed from the controller. But since we only have one day till January, I'd rather get this out the door than hyperfocus on cleaning up the codebase, especially since remaining optimizations don't affect functionality & we won't be developing this app extensively.

src/views/Home/index.view.tsx
mocklee: https://github.com/CruzHacks/cruzhacks-2021-website/pull/122#discussion_r550619673

mocklee: I went ahead & pulled out all homepage props into JSON imports from a `props` directory in https://github.com/CruzHacks/cruzhacks-2021-website/pull/122/commits/02437f72755ddf8099f3780b27656cfe3afcba52 — now the controller code is looking super-clean

ilopezro (approved):
Let's ship 🚀

PR #123, opened 2020-12-20
Minimal iteration on responsiveness

mocklee:
Problem
=======
Addressing responsiveness of buttons & MLH banner + introducing new background on minimal iteration

Screenshots (optional):
-----------------------
Screenshot from 2020-12-19 16-26-09
Screenshot from 2020-12-19 16-26-47


src/views/Home/Home.scss
```diff
@@ -1,18 +1,17 @@
+@use "~components/Button/Button";
+
```
ilopezro, line 2: This shouldn't be necessary if we import the scss file inside `components/Button/index.view.tsx`. If it's not there, we should import it in that file and remove this line.

src/views/Home/index.view.tsx
```diff
@@ -22,9 +21,40 @@ const description = [
...
+          <a
+            className="Hero__details--link"
+            href="mailto:contact@cruzhacks.com"
+            target="_blank"
+            rel="noreferrer"
+          >
+            Contact Us
+          </a>
+        </span>
+      </>
```
ilopezro, line 54: Is there a way to condense this code? To me it looks a bit repetitive & I think there can be a way to avoid adding an entire react component here.

ilopezro (changes requested):
The rest LGTM -- please take a look at my comments though. I think we can definitely refactor and condense some code.

src/views/Home/index.view.tsx
mocklee: yea you're right, I was a bit lazy here & didn't bother to conditionally render links at the hero component, where this code should be at. I went ahead & turned this back to simple strings

src/views/Home/Home.scss
mocklee: yep, turned out this wasn't necessary since the index for `buttons` was already packaged with `buttons.scss` & included in the `Home` index.

ilopezro: Ahh gotcha! I didn't see line 13. That's on me!

ilopezro (approved):
I think I will do a quick cleanup of the code. I see we have multiple ways of storing information within our `Home` view. We use both objects and arrays and I think we should try and be consistent. This looks good to go though!

PR #129, opened 2020-12-23
baseline FAQ implemented

PriyaRajarathinam:
Problem
=======
As a hacker, 
I want an FAQ section 
So some of my questions can be answered before contacting CH 
[Link to Asana Ticket](https://app.asana.com/0/1197738625416321/1199382769817325)

Solution
========
Created a FAQ component that answers some basic questions

Change summary:
---------------
* Created a baseline FAQ component
Dev-Ops
=======
- [ ] `.env` & `.env.sample` 
- [ ] `webpack.config.js` 
- [ ] GCP Secret Manager
- [ ] Github Secrets 
- [ ] GithubActions Workflow

Steps to Verify:
----------------
1. yarn start
2. View FAQ component

Screenshots (optional):
-----------------------

Screenshot


src/views/Home/components/FAQ/faq.scss
```diff
@@ -1,3 +1,106 @@
...
+  &__text--question {
+    @include typography.source-description-purple;
+
+    font-style: normal;
+    font-weight: bold;
+    color: var(--pink);
+    margin: 1em;
+  }
+
+  &__text--answer {
```
mocklee, line 54: add padding here to alleviate the hard cutoff here & add some negative space at the base of the component to frame the text better
Screenshot from 2020-12-25 18-55-46

mocklee (changes requested):
great work getting this off the ground. just had some nitpicks with styles, but this is close!

src/views/Home/components/FAQ/faq.scss
```diff
@@ -1,3 +1,106 @@
-/* stylelint-disable-line no-empty-source */
+@use "~stylesheets/typography";
+@use "~stylesheets/responsiveness";
+@import url("https://fonts.googleapis.com/css2?family=Source+Code+Pro&display=swap");
```
ilopezro, line 3: Not entirely sure why we'd need this? Don't we already have these fonts in from the schedule component?

src/views/Home/components/FAQ/index.view.tsx
```diff
@@ -1,8 +1,100 @@
 import React from "react";
```
ilopezro, line 1: #71 allows us to get rid of importing react if not needed; however, since we do use it, let's use this react import: 
`import * as React from "react"`. Future versions of react is deprecating the usage of this current import I believe.

mocklee (approved):
FAQ looks good— let's take this to staging!

src/views/Home/components/FAQ/index.view.tsx
```diff
@@ -1,8 +1,102 @@
...
+      question: "When will decisions be made?",
+      answer: "Decisions sent out a week before the hackathon.",
+    },
+  ],
+];
 const FAQComponent: React.FC = () => {
-  return <div className="FAQ">Hello from FAQComponent</div>;
+  return (
+    <div className="FAQ">
+      Hello from FAQComponent
```
ilopezro, line 62: Let's get rid of this for staging!

src/views/Home/index.view.tsx
```diff
@@ -99,6 +100,7 @@ const HomepageView: React.FC = () => {
             </div>
           ))}
         </Hero>
+        <FAQ />
```
ilopezro, line 103: `FAQ` should go after `Mission`

PR #130, opened 2020-12-25
footer component

tim-nguyen-cs:
Problem
=======
As a prospective hacker,
I would like to view additional footer resources
To resolve more inquiries and questions I have about CruzHacks.

[Link to Asana Ticket](https://app.asana.com/0/1197738625416321/1199382769817337/f)

Solution
========
To implement the Footer section of the homepage, flexbox-based positioning and margin alignments were used to replicate the provided Figma design with responsiveness. Although preliminary social media links and other graphics have not been included yet, this first draft outlines boilerplate code regarding how the footer is to be implemented.

Change summary:
---------------
* first iteration of footer https://github.com/CruzHacks/cruzhacks-2021-website/commit/f8fd13933a8fe9f65957f78fbe512cc41ffbf231

Steps to Verify:
----------------
1. Checkout branch using `git checkout feature/footer`
3. Verify that footer section is designed and responds correctly.


mocklee (commented):
looks good, just one general issue with the contrast here.. this is tied into the design, but maybe add `text-shadow` to add a bit more contrast
Screenshot from 2020-12-25 19-45-12

tim-nguyen-cs:
> looks good, just one general issue with the contrast here.. this is tied into the design, but maybe add `text-shadow` to add a bit more contrast
> Screenshot from 2020-12-25 19-45-12

Added `font-shadow` per your suggestions - it looks great! Thanks for the tip - let me know if there's anything else you want me to change!

src/views/Home/Home.scss
```diff
@@ -15,3 +15,7 @@
     margin-bottom: var(--medium-spacer);
   }
 }
+
+a {
+  color: inherit;
+}
```
ilopezro, line 21: Could we use class names instead of generic attributes? I understand why we have this here, but I don't see the `a` tag anywhere within `Home` index view. Let's give the `a` tags a classname within the `Footer` component and style the attributes there.

src/views/Home/components/Footer/index.view.tsx
```diff
@@ -0,0 +1,52 @@
+import React from "react";
```
ilopezro, line 1: Because we upgraded to a new version of typescript in #71, we no longer have to import `react` if we don't use it. Instead, we can import it using the following since this is the convention that it is moving towards using this newer version of react: 
`import * as React from "react";`

src/views/Home/components/Footer/index.view.tsx
```diff
@@ -0,0 +1,52 @@
...
+        <div>Links!</div>
+        <div>Scroll!</div>
+      </div>
+      <div className="Footer__text">
+        <div className="Footer__text--left">
+          Made with ❤️☕❤️ from Santa Cruz, California.
+        </div>
+        <div className="Footer__text--middle">
+          CruzHacks was formerly known as HackUCSC, and launched in April 2014.
+          The founders were&nbsp;
```
ilopezro, line 18: SUPER NIT: I'm honestly not that bit a fan of having `&nbsp;` directly in the codebase as it creates a bit of messy code. We can instead [style the `a` tag accordingly ](https://techstacker.com/how-to-add-white-space-between-elements-react-jsx/). I'd prefer if we styled it using SCSS instead of hardcoding it into the text like that.

ilopezro (commented):
I think this looks good overall! Please take a look at my comments & let me know what you think!

src/views/Home/index.view.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from "react";
 import Hero from "views/Home/components/Hero/index.view";
+import Footer from "views/Home/components/Footer/index.view";
```
ilopezro, line 3: One last thing, we don't need to import all the way up the tree from `views/` directory. We can simply import using `./components/Footer/index.view`

src/views/Home/Home.scss
tim-nguyen-cs: That's understandable - no problem!

src/views/Home/components/Footer/index.view.tsx
tim-nguyen-cs: Okay, cool ~ no problem! In that case, we might want to change this notation for all other files too!

src/views/Home/index.view.tsx
tim-nguyen-cs: Cool ~ I would recommend changing this in the previous line too in another pull request.

src/views/Home/components/Footer/index.view.tsx
tim-nguyen-cs: I agree that these entities make code look a little unclean, but I'm not quite sure what about the resolution that you mentioned. The `&nbsp;` entities are here, because our linter requires that content within `a` tags be placed on a new line. Although I personally prefer inline `a` tags, this causes sentences to be partitioned weirdly and requires `&nbsp;` used in place of any trailing spaces that are pre-maturely truncated.

Let me know if this makes sense and how to proceed - also note that this section will be removed later when this sentence is integrated into our `Teams` or `Sponsors` page.

src/views/Home/index.view.tsx
ilopezro: Since @mocklee 's PR (#122) deals with `Hero` component, I will leave a comment there to change line 2.

src/views/Home/components/Footer/index.view.tsx
ilopezro: Yeah -- I believe I ran a command that changed all the existing import statements. This only appears once or twice in the codebase. I did a `cmd + F` on the existing codebase and it looks like I added the wrong import statement after the PR was merged in. I think we should address those in a separate PR. Thoughts?

src/views/Home/components/Footer/index.view.tsx
ilopezro: Ahh yes! I see what the issue is. Since the `a` tag has 3 props and contents within it, in order to maintain the character count for that line to 120, it makes the `a` go to a separate line. Adding a space at the end of the line before the `a` tag would cause the linter to cry since we don't want trailing spaces. In order to remove these entities, we can [style it in the css](https://techstacker.com/how-to-add-white-space-between-elements-react-jsx/#or-just-use-css) by adding a `margin-left: .5rem` to the `a` tags.

src/views/Home/components/Footer/index.view.tsx
tim-nguyen-cs: Yep! I was lowkey planning a separate PR to just clean up the codebase a little bit, so that sounds like a good idea!

src/views/Home/index.view.tsx
tim-nguyen-cs: Sounds good!

src/views/Home/components/Footer/index.view.tsx
tim-nguyen-cs: Ohh, that sounds like a clever fix! I'll have that done by tonight!

mocklee (approved):
looks good from here! let's stage this up

ilopezro (commented):
Dope work! 🚀

tim-nguyen-cs:
I'll touch this up again when other parts of our webpage get flushed out, but let's get this in!

PR #144, opened 2021-01-03
Optimize copywriting (text content)

mocklee:
Problem
=======
I want to understand the mission, goals, and opportunities CruzHacks evangelizes. 

Solution
========
Color in some more details from the initial copywriting


src/views/Home/props/prizes.js
```diff
@@ -10,19 +10,19 @@ export const topics = [
     title: "Health Hacks",
     icon: <HealthHacksIcon />,
```
ilopezro, line 11: Yo this is actually super dope! Have a component in as a prop for the JSON object! Didn't catch this in the #97 PR.

ilopezro (approved):
This looks really good to me! 🚢

src/stylesheets/_typography.scss
```diff
@@ -1,4 +1,5 @@
 @use "~stylesheets/responsiveness";
+@use "~stylesheets/variables";
```
ilopezro, line 2: I'm wondering what this does though 🤔

mocklee: still cleaning up a bit!

PR #145, opened 2021-01-03
Update sponsor logos.

w-gao:
Problem
=======
I want to update the sponsors section to include our new sponsors. 

Solution
========
I added the EarthHacks and echoAR logos and rearranged the logos grid according to the newer Figma design.

Screenshots (optional):
-----------------------

image

Updated grid:
image


mocklee (approved):
let's update those logos!

PR #157, opened 2021-01-07
update Sticker Mule link

tim-nguyen-cs:
Problem & Solution
=======
* Updated Sticker Mule link, per Kelly's request on Slack.

### DIFF
--- a/src/views/Home/props/sponsors.js
+++ b/src/views/Home/props/sponsors.js
@@ -103,7 +103,7 @@ export const sponsors = [
     name: "Stickermule",
     className: "stickermule",
     logo: Stickermule,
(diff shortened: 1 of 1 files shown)


ilopezro (approved):
Not a *huge* deal, but make sure to provide a more thorough explanation for a PR. The PR template is not necessary because it's such a small change, but I shouldn't have to go digging through slack to see what or why we have this change in place. A simple

# Problem & Solution 
Problem & solution here

Should do the trick 👍

w-gao (approved):
@tim-nguyen-cs Thank you for the quick update!

tim-nguyen-cs:
Sure, I updated it now and will remember for the next time!

PR #158, opened 2021-01-07
Fix sponsors logo visibility on mobile.

w-gao:
Problem
=======
The SVG logos with embedded images are not showing correctly on mobile. 

Solution
========
Add back the width properties for the image tags. 

Steps to Verify:
-----------------------
- Checkout the `chore/fix-sponsors-svg-on-ios` branch.
- Verify that all logos are visible and in the correct place! **I've only tested on a few devices**.

Screenshots (optional):
-----------------------
Before:

After:


mocklee (approved):
thanks for this catch @w-gao, unfortunately I don't have an iOS device & we didn't consider expanding cross-browser testing`, which would've made this apparent earlier. 

this is pretty high priority, @ilopezro can you run one more deploy train so we can patch this on prod ASAP?

PR #160, opened 2021-01-08
Added colors and fixed overlap

PriyaRajarathinam:
Problem
=======
Text was overlapping and there were missing colors
[Link to Asana Ticket](https://app.asana.com/0/1197738625416321/1199333062665016)

Solution
========
Increased min-width and added colors.

Change summary:
---------------
* Fixed overlapping text and added colors
Dev-Ops
=======
- [ ] `.env` & `.env.sample` 
- [ ] `webpack.config.js` 
- [ ] GCP Secret Manager
- [ ] Github Secrets 
- [ ] GithubActions Workflow

Steps to Verify:
----------------
1. A setup step / beginning state
2. What to do next
3. Any other instructions
4. Expected behavior
5. Suggestions for testing

Screenshots (optional):
-----------------------


src/views/Home/components/Schedule/schedule.scss
```diff
@@ -119,21 +116,10 @@
...
-    margin-right: 20px;
-    left: calc(-2.3vw + 10px);
-    color: var(--pink);
-    transform: rotate(-90deg);
-  }
-
   &__title {
     @include typography.alt-title-text;
 
+    color: var(--salmon);
```
tim-nguyen-cs, line 122: Instead of adding `--salmon` here, can we change `--pink` to `--salmon` in the corresponding typography file?

tim-nguyen-cs (changes requested):
Looks good so far! Just one minor comment on where to change certain variable descriptions.

Do you think we can also update the heading to match that of "Keynote Speakers"? The title should be a little bolder if I remember.

src/views/Home/components/Schedule/schedule.scss
PriyaRajarathinam: Yeah I can do that! If I were to change salmon to pink would we have to change all the references to salmon?

tim-nguyen-cs: I think it should be fine if both are defined in `variables` and correlate to the same hex code. I think @ilopezro mentioned cleaning up the code later!

ilopezro:
This deepcode CI not running is super annoying. It's periodically decides it wants to stop working. I'm going to rebase with `main` so this branch is up to date, which'll cause a force reset on the CI job.

src/views/Home/components/Schedule/index.view.tsx
```diff
@@ -40,12 +64,12 @@ const ScheduleComponent: React.FC = () => {
           </div>
           <div className="Schedule__greyRect"></div>
           <div className="Schedule__eventContainer">
```
mocklee, line 66: good DOM structure here! to remove the last rule line from rendering in this list of schedule events, [use the `:last-child` selector](https://css-tricks.com/almanac/selectors/l/last-child/)

src/views/Home/components/Schedule/schedule.scss
```diff
@@ -12,26 +12,28 @@
...
-    margin: auto;
-    left: 25%;
-    bottom: 25%;
-    width: 50%;
-    min-width: 300px;
+    left: 50%;
+    margin-left: -32%;
+    bottom: 40%;
+    width: 75%;
+    min-width: 500px;
```
mocklee, line 19: I'd apply a relative unit to this min-width using `calc` so components remain elastic. 
Screenshot from 2021-01-13 18-15-51

src/views/Home/components/Schedule/index.view.tsx
```diff
@@ -40,12 +64,12 @@ const ScheduleComponent: React.FC = () => {
...
-            {scheduleItems[selectedDay].events.map(({ event, time }) => {
+            {scheduleItems[selectedDay].events.map(({ event, time, end }) => {
               return (
                 <div className="Schedule__event" key={event}>
                   <div className="Schedule__eventText">{event}</div>
                   <div className="Schedule__timeText">{time}</div>
-                  <div className="Schedule__greyRect--variant"></div>
+                  {renderDivider(end)}
```
mocklee, line 72: Interesting way to render a rule line; CSS could also make this easier with the `:last-child` & `:nth-child` selectors, but that would involve an event subcontainer for each time slot. In this case, [use the map index to conditionally render a rule line if we're at the last index of the event map](https://stackoverflow.com/questions/38364400/index-inside-map-function)

It's also a bit visually unclear if there's two workshops in a single time slot. I'd recommend leveraging the `end` variable you have to set the renderDivider function to provide a light-grey, dashed/dotted divider if there's another event in the same time slot

mocklee (changes requested):
Left a few notes for last-minute polishing— we're also missing two workshops on day 1. But it's cool to see our timetable come together, we're almost there with the schedule!

mocklee (approved):

PR #162, opened 2021-01-10
Added Remaining Speakers

ilopezro:
Problem & Solution 
=======
We did not have the complete lineup of speakers, so the workaround was having a `ComingSoon` component for the cards; however, we not have the final lineup ready and I've added the images and added the cards. 

Steps to Verify:
----------------
1. Checkout this branch and start a development environment 
2. Ensure that you see the speakers in alphabetical order based on Opening v Closing

Screenshots:
-----------------------
**Before**
Screen Shot 2021-01-09 at 5 14 04 PM

**After**
Screen Shot 2021-01-09 at 5 14 32 PM


tim-nguyen-cs (approved):
This looks good to me! One great thing about abstracting our data structure is that we can make easy changes like these relatively quickly!

I saw also you removed the fragment wrapper from the `Hero` component - although I'd think it'd be best to separate feature implementation from source-code clean-up, I think it's fine for now.

ilopezro:
Good point! I'll strip that out to keep this PR focused on implementing/finishing speaker section.

PR #163, opened 2021-01-10
Add new sponsors.

w-gao:
Problem
=======
I want to update the sponsors section to include our new sponsors.

Solution
========
I added the Santa Cruz Works, Progress, Blackstone and CROSS logos and rearranged the logos grid according to the newer Figma design.

Steps to Verify:
----------------
1. Checkout the `chore/update-sponsor-logos-v2` branch
2. Verify that the sponsor logos look good on all viewports
3. Also verify the links

Screenshots (optional):
-----------------------

image

image


ilopezro (changes requested):
This LGTM! Let's double check with Sponsorship in regards to links. Only one I am not entirely sure about is `Progress`. Also, can we get a [CROSS sponsor](https://cross.ucsc.edu/) for Kilo. I believe they also signed on. Again, double check with Sponsorship.

ilopezro (approved):
This LGTM! We will have scholarship/presidents verify logos on staging. The magic of having a staging environment 😁

w-gao:
Google Cloud needs to be added as well.

mocklee (changes requested):
good work on getting new sponsors integrated as they arrive! these logos are starting to look a bit crowded though— it'll work out better for us if we add some padding/margin to crowded logos by the 15th so they remain distinct & clickable.

w-gao:
@mocklee Let me know how this looks:

image

image

The logos might still be a little crowded on smaller devices but we don't have much room to play with.

mocklee:
> The logos might still be a little crowded on smaller devices but we don't have much room to play with.

We can optimize the space we have & leverage CSS grid to slip in more space. I applied a `row-gap` here to lend more padding on a row basis & it solves most of the crowding issue
Screenshot from 2021-01-13 19-51-47

mocklee (approved):

PR #172, opened 2021-01-14
Patch schedule max-width & scaling

mocklee:
Problem
=======
As a widescreen laptop user, I would like to view the schedule component without experiencing a hardcoded max width.
Screenshot from 2021-01-13 22-54-10
As a developer, I would like to have consistent scaling of all components on the homepage.

Solution
========
The schedule should depend more on relative than absolute positioning. Numerous CSS breakpoints are applied to scale the centered component + background across the entire viewport.  

Screenshots (optional):
-----------------------
Screenshot from 2021-01-13 22-55-45

### DIFF
--- a/src/views/Home/components/Schedule/schedule.scss
+++ b/src/views/Home/components/Schedule/schedule.scss
@@ -109,7 +109,7 @@
   &__dayThree {
     position: absolute;
     left: -5%;
-    top: calc(59% - 0.05vh);
+    top: calc(59.1%);
   }
 
   &__event {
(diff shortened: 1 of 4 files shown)


src/views/Home/components/Schedule/schedule.scss
```diff
@@ -84,19 +97,19 @@
...
     left: -5%;
-    top: 25.8%;
+    top: calc(25.8% - 0.05vh);
   }
 
   &__dayThree {
     position: absolute;
     left: -5%;
-    top: 59%;
+    top: calc(59.1%);
```
w-gao, line 112: Should this just be `59.1%`?

mocklee: yep 😅

ilopezro (approved):
Yo this looks really dope! Let's get this rolling 👍🏽

PR #173, opened 2021-01-14
Feature/footer

tim-nguyen-cs:
Problem
=======
Using `react-router-dom` prevents the `EmailSubscription` component from working properly because of an underlying HTML element.

Solution
========
Removed `react-router-dom` from the corresponding components and using `<a>` hypertext references instead. Also updated the navigation bar title to link to the homepage and revised the `Footer` component's property definitions.


w-gao (approved):
Looks good to me!

PR #176, opened 2021-01-15
patch dates for conditional schedule tabbing

mocklee:
Problem
=======
The default schedule tab is 1 day ahead of the actual date, resulting in the schedule rendering a page ahead on default.

Screenshot from 2021-01-15 00-53-43

Solution
========
Correct ternary conditions by -1 day, so the schedule knows what to render on the appropriate date.

Screenshot from 2021-01-15 00-53-54


tim-nguyen-cs (approved):
Nice patch!

PR #177, opened 2021-01-16
updated schedule

tim-nguyen-cs:
Problem & Solution
=======
Added new MLH workshops and revised text.


w-gao (approved):
Looks good!

This is starting to look crowded on mobile. We might need to adjust the height if we have more events to add.

PR #180, opened 2021-01-16
fixed line overlap

PriyaRajarathinam:
Problem
=======
What is the problem this work solves, including

[Link to Asana Ticket](https://app.asana.com/1234)

Solution
========

Created border instead of divider class

Change summary:
---------------
* Tidy, well formulated commit message
* Another great commit message
* Something else I/we did

Dev-Ops
=======
- [ ] `.env` & `.env.sample` 
- [ ] `webpack.config.js` 
- [ ] GCP Secret Manager
- [ ] Github Secrets 
- [ ] GithubActions Workflow

Steps to Verify:
----------------
1. A setup step / beginning state
2. What to do next
3. Any other instructions
4. Expected behavior
5. Suggestions for testing

Screenshots (optional):
-----------------------
image

### DIFF
--- a/src/views/Home/components/Schedule/schedule.scss
+++ b/src/views/Home/components/Schedule/schedule.scss
@@ -27,8 +27,8 @@
 
   &__schedule {
     position: relative;
-    width: 75%;
-    min-width: max(314.25px, 70vw);
+    width: 70vw;
+    min-width: 314.25px;
     max-width: 1000px;
     min-height: auto;
     height: fit-content;
--- a/src/views/Home/components/Schedule/schedule.scss
+++ b/src/views/Home/components/Schedule/schedule.scss
@@ -113,18 +113,18 @@
   }
 
   &__event {
-    width: 100%;
+    width: 97%;
     position: relative;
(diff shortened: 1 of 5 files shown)


tim-nguyen-cs:
@PriyaRajarathinam - Cool! It's looking good so far - I made a few comments below on some of my suggestions:

- [ ] Instead of using the dotted borders, do you think we can omit that to be a little more accurate to the Figma design?

- [ ] If we can touch up the styling to be a little lighter too, I think that could look a bit more appropriate.

- [ ] See if we can use something like `text-width` too to align the borders correctly also.

src/views/Home/components/Schedule/schedule.scss
```diff
@@ -28,7 +28,7 @@
   &__schedule {
     position: relative;
     width: 75%;
-    min-width: min(500px, 70vw);
+    min-width: max(314.25px, 70vw);
```
tim-nguyen-cs, line 31: Let's double-check this parameter on very large screen sizes!

image

tim-nguyen-cs (approved):

PR #234, opened 2021-03-30
Feature: Adding Functionality to access Organizer Applications

krishnakpandian:
Problem
=======
Users need to be able to easily access the CruzHacks 2022 Organizer Applications.

Solution
========
What I did to solve this problem was I added an additional button within the navbar dropdown menu linking to the organizer applications. I also added near the top of the main container of the main page another button redirecting users to the CruzHacks 2022 Organizer Applications

Change summary:
---------------
* Added functionality for organizer application forms in the navbar and the main container

Screenshots (optional):
-----------------------
Red Outline surrounding changes
Screen Shot 2021-04-01 at 11 10 39 PM

### DIFF
--- a/src/views/Home/index.view.tsx
+++ b/src/views/Home/index.view.tsx
@@ -12,6 +12,7 @@ import SpeakersComponent from "./components/Speakers/index.view";
 import FAQComponent from "./components/FAQ/index.view";
 import SponsorsComponent from "./components/Sponsors/index.view";
 import FooterComponent from "./components/Footer/index.view";
+import Button from "../../components/Button/index.view";
 
 import { navbarProps } from "./props/navbar.js";
 import * as heroProps from "./props/hero.json";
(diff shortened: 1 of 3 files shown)


src/views/Home/index.view.tsx
```diff
@@ -12,6 +12,7 @@ import SpeakersComponent from "./components/Speakers/index.view";
 import FAQComponent from "./components/FAQ/index.view";
 import SponsorsComponent from "./components/Sponsors/index.view";
 import FooterComponent from "./components/Footer/index.view";
+import Button from "../../components/Button/index.view";
```
ilopezro, line 15: Because our root folder is `src` we can reference everything from there. This can simply be: 
```suggestion
import Button from "./components/Button/index.view";
```

src/views/Home/props/organizerForm.js
```diff
@@ -0,0 +1,9 @@
+import { ButtonTypes } from "../../../components/Button/index.view";
```
ilopezro, line 1: Same here. We should reference everything from `src`. This will become 
```suggestion
import { ButtonTypes } from "./components/Button/index.view";
```

ilopezro (approved):
This looks good to me! Let's take a look at what it looks like on staging and then I will create a production deploy once everything checks out!

